PR #1, opened 2019-05-23
Develop build and install system

jeizenga:
I developed a functioning build system for the repository as it currently exists. It's based on make instead of cmake. My conversations with @adamnovak made me think that cmake external projects weren't exactly the most convenient tool for this task (although I'm sure we could have used them). Also, I know how to do make and git submodules, and I don't know how to do cmake. The middle-ground solution I came up with for now is to count on system-discoverable dependencies in this build, but to also have a separate repository with git submodules that coordinates all of the dependencies for you. Hopefully this will make it not too hard to integrate with VG later.


ekg:
I would prefer cmake. I suspect that we can learn the right way to use it
to work around the problems that Adam has found with it.

That tends to make it easier to package and integrate this into other
projects using cmake.

adamnovak:
@ekg If you want to cmake-ify this, please do it! That's the way I was heading when I started on it, but I didn't have a chance to work on it and @jeizenga did, and he's much more productive with Make. We figured a working build system in Make was better than a non-working one in Cmake, since we need to get something usable ASAP for @emikoifish to do experiments.

jeizenga:
Truth. I'm not at all attached to this system, but it works, which is an advantage over what we had before.

PR #20, opened 2019-09-06
add vectorizable overlay

glennhickey:
This is a quick and dirty overlay to make any graph vectorizable (ie packable).  

While testing this, I discovered that xg's implementation of the interface is inconsistent in that `node_vector_offset()` returns a 0-based offset, but `node_at_vector_offest()` expects a 1-based offset.  This should probably get fixed up, but will wait for xg to stabilize first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ INSTALL_PREFIX?=/usr/local
 INSTALL_LIB_DIR=$(INSTALL_PREFIX)/lib
 INSTALL_INC_DIR=$(INSTALL_PREFIX)/include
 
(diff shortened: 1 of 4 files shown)


ekg:
I also noticed the issue about 1-based vs 0-based indexes for these two functions. 

I think 0-based is nicer. We should stabilize on that unless you object.

I'd like to merge the nearly-there xg integration before fixing this though.

jeizenga:
+1 to 0-based indexing

glennhickey:
Sure, 0 works for me!

PR #25, opened 2019-09-24
Memory efficient path position overlay

jeizenga:
@glennhickey 

Adds a memory-efficient, static overlay that adds the `PathPositionHandleGraph` interface over a generic `PathHandleGraph`. 

Also adds a new external dependency: `BBHash`. They have a somewhat odd software design where the master branch provides hashing for (as far as I can tell) only integer types. If you want to hash non-integer types you have to use a separate branch `alltypes` in the same repository, so the dependency is actually on that branch.


ekg:
Cool! Much better than std::maps!!!!

PR #26, opened 2019-09-25
Make vectorizable overlay smaller

glennhickey:
Win inspiration from #25, use bbhash instead of normal hash tables for keeping track of node and edge ranks in the vectorizable overlay.  

I'm using parallel construction for the bbhash, and took the liberty of activating this for the PackedPositionOverlay.  @jeizenga is this okay?  libbdsg's already dependent on omp, and I think we'll generally be building overlays at the top level where we have access to all the threads.

### DIFF
--- a/include/bdsg/utility.hpp
+++ b/include/bdsg/utility.hpp
@@ -5,6 +5,7 @@
 #include <sstream>
 #include <iomanip>
 #include <functional>
+#include <omp.h>
 
 namespace bdsg {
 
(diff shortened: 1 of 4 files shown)


jeizenga:
Yeah, I wasn't sure how to handle that in my case. I did one thread, but I take your point. We'll probably only be using these overlays on large reference graphs.

ekg:
FWIW I use them for visualization on all kinds of graphs

jeizenga:
Hmm. Well, I suppose the real question is whether we'll be using them on lots of graphs simultaneously rather than on their size. I imagine you're usually visualizing one graph at a time in one process?

ekg:
Visualizing or binning for analysis in R. I've actually been building this
overlay ad hoc, so it'd be better to have something stable and upstream to
work with.

PR #30, opened 2019-11-21
add (but not implement) increment_node_ids method

glennhickey:


glennhickey:
@jeizenga would you mind giving packed/hash graph a shot whenever you have a chance?  thanks!

PR #32, opened 2019-12-04
Implement ID increment for HashGraph

adamnovak:
I'm doing it as an overlay, basically. We keep an offset between real IDs and internal IDs in the graph hashtable, translate dynamically when converting between handles and IDs, and only actually apply the offset to node IDs and path ID references when we serialize the graph.


adamnovak:
This implements/adds stubs for a couple libhandlegraph methods, but shouldn't actually depend on the libhandlegraph update that requires them. So I'm going to merge it.

PR #34, opened 2019-12-10
add cmake build process, ending need for libbdsg-easy

ekg:
:boom:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,8 @@
+\#*
+.#*
+*~
+build/
+bin/
+lib/
+test/
+.gdb_history
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,216 @@
+# Specify the minimum version for CMake
+
+cmake_minimum_required(VERSION 3.1)
+
+# Project's name
+project(bdsg)
+# We build using c++14
+set(CMAKE_CXX_STANDARD 14)
+
+# Use all standard-compliant optimizations
+set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3 -g")
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -g")
+
+# Use openmp for parallelism, but it's configured differently on OSX
+find_package(OpenMP)
+if (OPENMP_FOUND)
+  if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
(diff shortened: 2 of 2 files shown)


ekg:
@jeizenga @adamnovak this should eliminate the need for the second "libbdsg-easy" repository. Let's rebase around this change.

glennhickey:
Can we use this to get the real odgi into libbdsg (if it isn't already)?

ekg:
Yeah, it could be a good idea. I'm kind of uneasy about nesting libbdsg in
odgi.

glennhickey:
I was thinking about just using cmake to make your odji repo a submodule of
libbdsg.

adamnovak:
@ekg It looks like this is unconditionally getting and building the specified commits of the dependencies, even if they are already available in the environment, right?

How exactly do you see this working with vg's build system? Are we going to get libhandlegraph (or any other shared dependency) for the whole vg build from the external project inside of libbdsg? Is libbdsg going to build a libhandlegraph and vg is going to build another one and we just make sure they are always the same commit? Are we just going to keep building the old way in vg but port anything that would have used libbdsg-easy over to this?

This is the same style of build we are using in the xg repo, and we never properly integrated that into vg's build. vg is still just grabbing the source file and header file and building them itself.

jeizenga:
FWIW, whatever the best build process is, it makes a lot more sense to me to have the graph implementations consolidated in one repository (including XG and ODGI). Especially if we're planning on publishing all of them in one paper, I think they will be way more discoverable and less confusing for users if they're in one place.

jeizenga:
@adamnovak Would it be possible to arrange things so that VG doesn't have its own dependency on libhandlegraph and instead only uses libbdsg's version?

adamnovak:
Maybe. We'd have to figure out where cmake puts it and either point
our include paths into there or copy the headers and binaries out. It
would make working on libhandlegraph and building vg with the modified
version pretty tough, though.

jeizenga:
Also, I'm not sure where this OpenMP cmake configuration keeps getting copied from, but this is not the version that works on Mac (I get that this is hard to check without a Mac system handy). Take a look at xg or mmmultimap for the (uglier) configuration for OpenMP that compiles on Mac: 

https://github.com/ekg/mmmultimap/blob/master/CMakeLists.txt#L14-L35

ekg:
Ok thanks

Issue #24, opened 2019-09-18
Add magic numbers to the serialize/deserialize methods


adamnovak:
It would be handy to be able to identify the files' types. Serialize would prefix each file with some constant bytes for each graph type, and deserialize would skip over them.

We could also maybe add a method to SerializableHandleGraph that would return the magic number that the class uses, to let larger programs sniff types auto-magically.

VG's IO system now has support for ID-ing non-VPKG files by magic number, so this would help compatibility with vg.

jeizenga:
We could use one of those `_impl` schemes for `serialize`/`deserialize` to ensure that each implementation prefixes a serialization with its magic number.

jeizenga:
Proposal:

```
virtual static uint64_t get_magic_number();

virtual void serialize_impl(ostream& out) const;

virtual void deserialize_impl(istream& in);

virtual void serialize(ostream& out) const {
    out << htonl(get_magic_number());
    serialize_impl(out);
}

virtual void deserialize(istream& in) {
    uint64_t magic_number;
    in.read((char*) &magic_number, sizeof(magic_number) / sizeof(char));
    magic_number = ntohl(magic_number);
    assert(magic_number == get_magic_number());
    deserialize_impl(in);
}
```

PR #38, opened 2019-12-13
A magic number scheme for serialization

jeizenga:
@adamnovak I implemented the magic number scheme I proposed in https://github.com/vgteam/libbdsg/issues/24. Is this still something we want?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,7 +107,7 @@ set(hopscotch_map_INCLUDE "${INSTALL_DIR}/src/hopscotch_map/include")
 # libhandlegraph (full build using its cmake config)
 ExternalProject_Add(handlegraph
   GIT_REPOSITORY "https://github.com/vgteam/libhandlegraph.git"
-  GIT_TAG "4592bae22653e3a3c8666884f51490841803671f"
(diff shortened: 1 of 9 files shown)


adamnovak:
I don't think the `_impl` names are quite right for this, because `serialize` is going to apply the magic number and `serialize_impl` won't; the semantics are different and one is not an implementation of the other.

Maybe call it `serialize` and `write_members` or something?  But this is definitely the approach we want to take, I think.

ekg:
Ideally, serialize writes the magic number and load checks if it's the
expected number and barfs if not.

Other than that, I'm happy with anything that just depends on the single
magic number.

jeizenga:
@ekg That's pretty much what happens, but the implementation is in the public-scoped `serialize` method in libhandlegraph. 

@adamnovak I have no real attachment to the naming, so I can change it if you want. The idea was that the impl methods would be hidden by scope though, since they're protected in libhandlegraph.

adamnovak:
Yeah, only the `serialize`/`deserialize` methods should be public. But
the internal method shouldn't be whatever_impl unless it's really
*just* and implementation, I think.

PR #42, opened 2020-01-13
switch from externalproject_add of git repos to source_dir of submodules

ekg:
After this, we'll need to clone with --recursive now! But this should simplify builds, and removes the internet access dependency from them.


ekg:
This seems to have sorted things out. I'm going to merge this in, because it's helpful for packaging. I'll need this when integrating libbdsg into odgi.

glennhickey:
As a [fan of libbdsg-easy](https://github.com/vgteam/libbdsg-easy/pull/5#issuecomment-615395389) I like this direction.  But don't forget it will require a bunch of compensatory changes in vg's build and submodule structure.

ekg:
I guess it will.

We should probably be building vg with cmake as well.

adamnovak:
@ekg Did you try `./make_and_run_binder.py` with the submodule structure? Because I'm trying to run it now on a branch with master merged in, and it can't find `BooPHF.h`. It used to be in `build/bbhash-prefix/src/bbhash-build/`, but now (because of the submodule source for the external project?) that directory is empty.

Does it work for you with this merged? (And if not, can we think of a way we can add running it to CI, given that it needs its own cached build of Clang?)

adamnovak:
OK, I think I have the binding generation working again in https://github.com/vgteam/libbdsg/pull/67.

PR #44, opened 2020-01-27
Add Cecilia's Python bindings

adamnovak:
This adds the Python bindings that Cecilia wrote, and the scripts to regenerate them.

I think the scripts may not do some of the bindings postprocessing I had to do manually (related to the binding code wanting to include non-standard files from the STL it actually happened to be running on), so that may need to be added later.


adamnovak:
@ctcisar I don't know where your PR for this went, but this is the right branch with your latest work here, right?

make_and_run_binder.py
```diff
@@ -0,0 +1,155 @@
...
+    # download the most recent copy of binder from git
+    if not glob.glob("binder"):
+        print("Binder not found, cloning repo...")
+        os.mkdir("binder")
+        git.Git(".").clone("https://github.com/RosettaCommons/binder.git")
+
+def build_binder():
+    # check for binder executable in the location we expect it
+    # if it's not there, build binder with the included script
+    # returns location of executable
```
adamnovak: I feel like these ought to be docstrings.

make_and_run_binder.py
```diff
@@ -0,0 +1,155 @@
...
+                     glob.glob(f'{this_project_source}/../include/**/*.hpp', recursive=True) + # this format needed to reach headers
+                     glob.glob(f'{this_project_source}/../include/**/*.cpp', recursive=True) +
+                     glob.glob(f'{this_project_source}/../include/**/*.h', recursive=True) +
+                     glob.glob(f'{this_project_source}/../include/**/*.cc', recursive=True) +
+                     glob.glob(f'{this_project_source}/../include/**/*.c', recursive=True) +
+                     glob.glob(f'{this_project_source}/../build/*/src/*/*.hpp', recursive=True) + 
+                     glob.glob(f'{this_project_source}/../build/*/src/*/*.cpp', recursive=True) + 
+                     glob.glob(f'{this_project_source}/../build/*/src/*/*.h', recursive=True) + 
+                     glob.glob(f'{this_project_source}/../build/*/src/*/*.cc', recursive=True) + 
+                     glob.glob(f'{this_project_source}/../build/*/src/*/*.c', recursive=True)):
```
adamnovak: I'm sure this could be condensed to a collection of extensions and maybe `os.walk`.

make_and_run_binder.py
```diff
@@ -0,0 +1,155 @@
...
+    os.mkdir(bindings_dir)
+    # be sure to include dirs for dependencies so it won't get lost
+    proj_include = glob.glob("build/*/src/*/include")
+    proj_include = " -I".join(proj_include)
+    command = (f'{binder_executable} --root-module {python_module_name} '
+               f'--prefix {os.getcwd()}/{bindings_dir}/ '
+               f'--bind {this_project_namespace_to_bind} '
+               + ('--config config.cfg ') +
+               f' {all_includes_fn} -- -std=c++14 '
+               f'-I{this_project_include} {proj_include} -DNDEBUG -v').split()
```
adamnovak: It's not a good idea to build a big string and then split here, in case of spaces in arguments. This should just be built as a list to begin with.

ctcisar:
> @ctcisar I don't know where your PR for this went, but this is the right branch with your latest work here, right?

Yes, this is the right branch.  I'll work on integrating these suggestions this afternoon.

make_and_run_binder.py
ctcisar: I didn't know this was an option-- I'll look into it.

ctcisar:
Updated to docstrings and changed the formatting of the binder command to be a list, not a concatenated string.  I tried to implement a different method of traversing the project files (included in comments), but the #include replacement function seems to not be working, which is obscuring whether that traversal is right.

adamnovak:
That's weird. It worked before, right?

ctcisar:
I thought so, yes.  At this point I'm starting to doubt my own recall though..

PR #47, opened 2020-03-17
Improve binding build and get bindings up to date

adamnovak:
I've fixed up the binding build process so it runs on Ubuntu 18.04, and updated the bindings to match the current source, with the changes to serialization that broke the old bindings accounted for.

When I generate the bindings on my machine, they don't appear to attempt to include any internal STL headers, so I have some hopes for these working out of the box on Mac. However, I haven't tested them yet.


jeizenga:
Nice. Thanks for pushing on this so quickly.

adamnovak:
OK, these now build on Mac, but don't know how to find libhandlegraph.dylib. Also, `bind_bdsg_odgi` doesn't make it into the module library somehow.

adamnovak:
OK, this now builds on Mac! It even works, as long as you `cp build/handlegraph-prefix/src/handlegraph-build/libhandlegraph.dylib lib/` after the build is finished.

I still have to work out why it doesn't know where to find the libhandlegraph dylib (maybe something fancy with Mac RPATH logic and install names?) but we should be able to merge this.

PR #55, opened 2020-03-31
Tighten semantics

adamnovak:
This makes sure that deleting nodes also deletes any paths that touch them, because that is at least a sensible, defined behavior.

It also prohibits empty nodes for everything except HashGraph. Some vg algorithms in the core of mpmap (like `extract_extending_graph`) expect to be able to create empty nodes (to represent sides that project into a subgraph) and they use HashGraph.

Clearly we should do something about that, but I'm not sure what yet.


jeizenga:
I think ODGI already has a logic for handling deletions of nodes on paths based on "hidden handles", right @ekg  ?

adamnovak:
OK, I added all the fixes to get CI to pass, and also a merge of #62, onto here. I'm going to merge this, and CI should work from here on out.

Issue #19, opened 2019-08-24
The ODGI graph does not pass the handle graph tests ported over from VG


jeizenga:
I haven't looked into this deeply yet, but we should figure out whether this is the fault of the tests or the implementation, or perhaps a disagreement over the generality of our formalism.

ekg:
Odgi test calls a version of these tests from a few months ago. What failed?

jeizenga:
Okay, I think I at least have a clue what's happening. It relates to this line in the orginal `node.cpp`:
https://github.com/vgteam/odgi/blob/master/src/node.cpp#L36

I think I wrote to you about this on slack a while back. The issue is that the assert statement was never actually executing because of this line in the CMakeLists, which removes all asserts
https://github.com/vgteam/odgi/blob/master/CMakeLists.txt#L235

I had to edit the line to have `edge_count()` instead of `edge_count` in order to compile. However, this assert statement has probably never actually run in the tests in the odgi repository because of that cmake flag. According to the slack messages I left you, you should be able to reproduce this from the `vg view` GFA from `tiny.vg`.

ekg:
I get it now. I forgot that release removed asserts.

This assertion is incorrect and should be removed or changed. The length of
res should be edge_count*EDGE_RECORD_LENGTH.

jeizenga:
Another update: if I comment out that assert everything passes. Is that assert statement correct?

ekg:
No, as I described, the assert is not correct. It was left over from a
previous version. It just needs to be removed.

ekg:
Ok I'll take a look. How do I get to the same state as you?

PR #58, opened 2020-04-03
Don't make the rank of the first node -1

adamnovak:
In #54 we replaced a variable that used to start at 1 with one that now starts at 0, so we need to not subtract 1 from it to get a 0-based rank.

### DIFF
--- a/include/bdsg/odgi.hpp
+++ b/include/bdsg/odgi.hpp
@@ -486,10 +486,10 @@ private:
     void set_handle_sequence(const handle_t& handle, const std::string& seq);
 
     /// get the backing node rank for a given node id
-    uint64_t get_node_rank(const nid_t& node_id) const;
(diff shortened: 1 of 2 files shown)


adamnovak:
This grew slightly, because the graph round trip test I was trying to get working kept shaking out more bugs, but I think I have it. This should be what I need to make https://github.com/vgteam/vg/pull/2702 pass.

jeizenga:
think you might have fixed https://github.com/vgteam/libbdsg/issues/19 ?

jeizenga:
Hey, actually just FYI, I just checked and it turns out that ODGI is now failing tests in `test_libbdsg::test_deletable_handle_graphs` that it used to pass. Womp womp.

jeizenga:
Not necessarily. A couple tests are commented out disabled on ODGI right now.

PR #70, opened 2020-06-04
Expand installation and C++ API documentation

adamnovak:
This adds more installation documentation, including how to install dependencies for more systems and how to deal with common build problems.

It also brings over the basic structure of the Python API reference page to the C++ version.


adamnovak:
This also now includes a bunch of fixes to make libbdsg ODGI file-compatible with the actual odgi tool.

adamnovak:
I'm still going to merge this, even though we're ditching https://github.com/vgteam/odgi/pull/125 in favor of more focused single-topic backports, because we need at least some of this for libbdsg ODGI to pass the libbdsg unit tests as a MutablePathDeletableHandleGraph.

PR #74, opened 2020-08-10
Work as an external project or add_subdirectory directory

adamnovak:
This has the changes to the CMake build system I needed to make this project a well-behaved external project or subdirectory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,8 +58,7 @@ elseif (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 endif()
 
 # Set the output folder where your program will be created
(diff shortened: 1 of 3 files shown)


adamnovak:
This should be enough to pass tests on Mac and Linux and to work as an `add_subdirectory()` in vgteam/libbdsg_example on Linux when installing the depending project.

I think to work on Mac with libbdsg_example when installing the depending project, we need to change the `external_project()` calls in this repo that result in dylibs into `add_subdirectory()` calls, so the actual final install step can be in charge of settign their install names. But I think that will be another PR.

adamnovak:
This should *work* as an external project too but whatever uses it that way would have the same install name problem on Mac that we have with our own dependencies.

PR #75, opened 2020-08-11
Upgrade DYNAMIC and convert to use add_subdirectory

adamnovak:


adamnovak:
Now we will use `add_subdirectory` and/or our own CMake targets for all our dependencies (except Binder, which I was afraid to touch and nobody transitively depends on), instead of external projects.

This means we can be used with `add_subdirectory` and properly install all our libraries to the right place where the very top level project is installing. It also means our configuration info is available to depending projects at config time.

PR #87, opened 2020-10-06
handle reversing self edges in divide_handle in odgi

subwaystation:
This should fix https://github.com/vgteam/vg/issues/3020.


subwaystation:
@jeizenga I think this is ready to merge.

Are there any further steps necessary to bring this into `vg`?

PR #91, opened 2020-12-11
Add and test support for memory-mapping serialization

adamnovak:


adamnovak:
I'm not quite sold on the current backend implementation and the interface it presents for inheritors. We also could simplify the code substantially if we pulled in `mio` to actually call mmap (at the cost of having to do something clever to get memory when there's no attached file). But I'm going to finally merge this.

PR #107, opened 2021-05-24
Actually pack the memory-mapped int vectors

adamnovak:
This makes our memory-mappable int vectors actually packed to the assigned bit width.

Unfortunately I think the Mac CI environment upgraded its Python version, causing our tests to start failing, so we will want to fix that before merging this.


adamnovak:
This doesn't quite work on my Mac yet; it complains we're trying to point across memory mapping "chains" when the tests run. I will need to investigate more.

adamnovak:
#109 replaced this I think.

Issue #116, opened 2021-08-18
When running on (M1?) Mac, tests fail due to trying to refer across chains


adamnovak:
After updating the submodules to versions that support M1 Mac and its Clang, when you build and run the tests, `test_paged_vector<PagedVector<5, MappedBackend>>();` fails with:

```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Attempted to refer across chains!
```

It looks like the problem is in the assignment operator for the `yomo::Pointer<>`, and it's trying to point from normal memory into a `yomo` memory chain. I tried hacking it to allow `yomo::Pointer<>` to work from non-`yomo` memory, but that led to a different failure further along in the test, so I think that really we have a mistake somewhere which GCC/Linux doesn't notice.

PR #121, opened 2021-08-23
Fix Github Actions YAML

adamnovak:
I got the syntax wrong the first time.


adamnovak:
Looks like our old Travis tests don't work anymore.

This runs at least the libbdsg C++ tests on Mac, which should avoid new problems like #116. We still need to bring back the Python doctests, and also test on Linux. And fix the ReadTheDocs build.

PR #122, opened 2021-09-03
Fix MacOS empty CMAKE_INSTALL_LIBDIR when using CMAKE_INSTALL_PREFIX variable

rlorigro:
This fixes the issue where linking libbdsg doesn't work on MacOS because the dylibs have the wrong path for their `INSTALL_NAME_DIR` or `LC_ID_DYLIB`.


jeizenga:
Fuck dylibs, all my homies link statically

rlorigro:
@jeizenga amen brother, but it seems the BUILD_STATIC flag has no effect when using CMake

rlorigro:
inb4 "fuck cmake"

jeizenga:
fuck cmake

rlorigro:
lets be real, if theres anyone we should direct our hate towards it's Apple

PR #124, opened 2021-09-10
Trim unused memory and add memory reporting

adamnovak:
This adds automatic collection of trailing free blocks in file-backed `UniqueMappedPointer`s when they are closed.

It also adds a `get_usage()` method to the pointer itself, which returns total bytes mapped, number of bytes free (including allocator headers), and number of bytes reclaimable from the end (including allocator headers). So you can see how efficiently a data structure has been packed by the allocator. Wasted space fraction would be ((free - reclaimable) / total), assuming you save via mapped file.

I've added/implemented some `shrink_to_fit()` methods, but those might actually make total file size worse by allocating a new smaller block from the end and leaving a big free hole where the old larger allocation was. If it does look like we need to pack things more tightly, I can implement some smarter reallocators/repackers that try to move things to the front of the file.

@xchang1 Can you use this?

### DIFF
--- a/bdsg/include/bdsg/internal/mapped_structs.hpp
+++ b/bdsg/include/bdsg/internal/mapped_structs.hpp
@@ -307,10 +307,17 @@ public:
     static void* find_first_allocation(chainid_t chain, size_t bytes);
     
     /**
-     * Get the fraction of bytes in the chain which are free, or NaN if the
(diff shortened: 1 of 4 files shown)


xchang1:
Yeah, this looks perfect. Thanks, Adam!

PR #127, opened 2021-09-16
Check chains and distance index

adamnovak:
This has @xchang1's new distance index, as well as #124, #126, and a fix for loading chains from streams so we don't split allocations across links and crash anymore.


adamnovak:
Looks like changing the load function's argument to const means I have to rebuild the bindings.

PR #128, opened 2021-09-28
Fast mem map

adamnovak:
This adds a bunch of acceleration to the `bdsg::yomo::Pointer` and `bdsg::yomo::UniqueMappedPointer`. Now, the `UniqueMappedPointer` no longer needs to lock anything to find its value, as the mapped memory address is cached. Also, the `Pointer` will no longer need to lock anything or consult the chain data structures, after the first access, if the in-chain offset also applies in memory. This can be guaranteed by unmapping and remapping the file, as files that are loaded are always loaded contiguously.

If we still need to make construction of things faster, we can try asking the OS for mappings that abut when we create them.


adamnovak:
I also dropped `big_endian`, and the old `MmapBackend` that turned out to not be thr way I went about implementing memory mapping.

Because of the removal of `big_endian` and the changed to `Pointer`, saved memory-mappable objects will have to be regenerated.

PR #131, opened 2021-11-15
Fix non-optimized build

adamnovak:
@xchang1 said that libbdsg wouldn't work without optimization. This fixes a constant that was missing a compilation unit and lets you use `-DOPTIMIZE=OFF` on the CMake command line to build without optimization.


adamnovak:
The un-optimized tests still pass, albeit quite slowly.

PR #153, opened 2022-03-30
Path metadata API implementation

adamnovak:
This adds implementations and bindings for the path metadata API in https://github.com/vgteam/libhandlegraph/pull/80


adamnovak:
This is for libhandlegraph with https://github.com/vgteam/libhandlegraph/pull/84

PR #165, opened 2022-09-09
Distance index

xchang1:
Update the distance index to be more efficient. This changes the format of the index so distance indexes (version 2) will have to be rebuilt
This only affects the distance index


xchang1:
@adamnovak I changed a bunch of config files and stuff that I don't completely understand. Can you double check that I didn't mess anything up please?

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+    
+    /*Root record
+    
+     * - The (single) root vector has the format:
+     *   [root tag, # connected components (N), # nodes (M), min_node_id, max depth [pointer to node/snarl/chain record] x N], [pointer to node+ node offset] x M]
+     *   The root vector stores the root of every connected component, which can be a 
+     *   node, snarl, or chain
+     */
+    const static size_t ROOT_RECORD_SIZE = 5;
+    const static size_t COMPONENT_COUNT_OFFSET = 1;
```
adamnovak, line 816: Should the offsets maybe be defined as the previous offset plus 1, and the size as the last offset plus 1, to make modification easier later, and to make sure none of the values can be wrong?

You might also be able to get away with making each of these "types" actually an enum, to let the compiler count for you.

Something like:
```
enum root_record_fields : int {
    ROOT_RECORD_TAG=0,
    COMPONENT_COUNT_OFFSET,
    NODE_COUNT_OFFSET,
    MIN_NODE_ID_OFFSET,
    MAX_TREE_DEPTH_OFFSET,
    ROOT_RECORD_SIZE
};
```

Also also, are you sure you don't want to consistently prefix all these names with the name of the type they belong to? If you see a line of code that says `record_start + NODE_COUNT_OFFSET`, it's not obvious that it really was supposed to be `record_start + SNARL_NODE_COUNT_OFFSET`.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+    const static size_t CHAIN_NODE_COUNT_OFFSET = 1;
+    const static size_t CHAIN_PARENT_OFFSET = 2;
+    const static size_t CHAIN_MIN_LENGTH_OFFSET = 3;     
+    const static size_t CHAIN_MAX_LENGTH_OFFSET = 4;
+    const static size_t CHAIN_RANK_OFFSET = 5;
+    const static size_t CHAIN_START_NODE_OFFSET = 6;
+    const static size_t CHAIN_END_NODE_OFFSET = 7;
+    const static size_t CHAIN_LAST_CHILD_OFFSET = 8; //The offset of the last thing in the chain - node or (if looping chain) snarl
+    const static size_t CHAIN_DEPTH_OFFSET = 9;
+    //DIstances from the left/right of the chain to the left/right of its parent
```
adamnovak, line 860:
```suggestion
    //Distances from the left/right of the chain to the left/right of its parent
```

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+    //As with other distances, store 0 for inf, value +1 for everything else
+    const static size_t NODE_DISTANCE_LEFT_START_OFFSET = 5;
+    const static size_t NODE_DISTANCE_RIGHT_START_OFFSET = 6;
+    const static size_t NODE_DISTANCE_LEFT_END_OFFSET = 7;
+    const static size_t NODE_DISTANCE_RIGHT_END_OFFSET = 8;
+ 
+    /*Chain record
+
+     * - A chain record for each chain, which is interspersed node and snarl records:
+     *   The nodes are all stored in a "TrivialSnarl", which is a bunch of nodes with no snarls between them
```
adamnovak, line 841: Maybe the TrivialSnarl should be a TrivialSnarlRun or something then?

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+     *   [trivial snarl tag, pointer to parent, node count, prefix sum, fd loop, rev loop, component]
+
+     * The record is followed by [node id+orientation, right prefix sum] for each node in the trivial snarl
+     * So the total length of the trivial snarl is 8+2*#nodes
+     * The right prefix sum is the sum from the start of the trivial chain to the right side of the node (relative to the chain)
+     * The node_record_offset in a net_handle_t to a trivial snarl points to a node in the trivial snarl
+ 
+     */
+    const static size_t BITS_FOR_TRIVIAL_NODE_OFFSET = 8;
+    const static size_t MAX_TRIVIAL_SNARL_NODE_COUNT =  (1 << BITS_FOR_TRIVIAL_NODE_OFFSET) -1;
```
adamnovak, line 878: Isn't this 255 and not 128 as described in the comment?

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+     *   inner node side of the start and end
+     *   Node count is the number of nodes, not including boundary nodes
+     */
+    const static size_t SNARL_RECORD_SIZE = 8;
+    const static size_t SNARL_NODE_COUNT_OFFSET = 1;
+    const static size_t SNARL_PARENT_OFFSET = 2;
+    const static size_t SNARL_MIN_LENGTH_OFFSET = 3;
+    const static size_t SNARL_MAX_LENGTH_OFFSET = 4;
+    const static size_t SNARL_DISTANCE_START_START_OFFSET = 5;
+    const static size_t SNARL_DISTANCE_END_END_OFFSET = 6;
```
adamnovak, line 907: These aren't described in the structure in the comment.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+    const static size_t SNARL_CHILD_RECORD_OFFSET = 7;
+
+    /*A simple snarl for bubbles with only nodes with two edges, one to each bound
```
adamnovak, line 910: How is this different from a trivial snarl then?

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+     * [simple snarl tag, node count+length, parent, [node id, node length]xN
+    */
+    const static size_t SIMPLE_SNARL_RECORD_SIZE = 3;
```
adamnovak, line 913: Maybe the "records" should really be "headers", since they structurally contain variable-length things that aren't included in their "size".

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+    //This one stores the node count and min and max lengths of the snarl
+    //It'll take 26 bits: 11 for each length
+    const static size_t SIMPLE_SNARL_NODE_COUNT_AND_LENGTHS_OFFSET = 1;
+    const static size_t SIMPLE_SNARL_PARENT_OFFSET = 2;
+
+     /*  At the end is the (single) child vector, listing children in snarls
+     *   [child vector tag, (pointer to records) x N
+     *   Each snarl will have a pointer into here, and will also know how many children it has
+     */
```
adamnovak, line 922: Why not include the lists of children with the things that have them?

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+                     "TRIVIAL_SNARL", "DISTANCED_TRIVIAL_SNARL",
+                     "SNARL", "DISTANCED_SNARL", "SIMPLE_SNARL", "OVERSIZED_SNARL", 
+                     "ROOT_SNARL", "DISTANCED_ROOT_SNARL",
+                     "CHAIN", "DISTANCED_CHAIN", "MULTICOMPONENT_CHAIN",
+                     "CHILDREN"};
+    vector<std::string> connectivity_t_as_string = { "START_START", "START_END", "START_TIP", 
+                            "END_START", "END_END", "END_TIP", 
+                            "TIP_START", "TIP_END", "TIP_TIP"};
+    vector<std::string> net_handle_record_t_string = {"ROOT_HANDLE", "NODE_HANDLE", "SNARL_HANDLE", 
+                                                "CHAIN_HANDLE", "SENTINEL_HANDLE"};
```
adamnovak, line 936: These should probably be `static const`. They could also just be `static const char*[]` or something, instead of full heap-allocated vectors and strings.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+
+
+    /* If this is 0, then don't store distances
+     * Otherwise, for snarls with more children than snarl_size_limit, only store the distances
+     * that include boundary nodes (OVERSIZED_SNARL)
+     */
+    size_t snarl_size_limit = 1000;
+    uint32_t magic_number = 1738636486;
```
adamnovak, line 944:
```suggestion
    static const uint32_t magic_number = 1738636486;
```

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -787,10 +785,185 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+     * The bit vector will be the last 6 bits of the tag
+     * 
+     * Each bit represents one type of connectivity:
+     * start-start, start-end, start-tip, end-end, end-tip, tip-tip
+     * 
+     * The remainder of the tag will be the record_t of the record
+     */
     /////////// Methods for interpreting the tags for each snarl tree record
 
     const static record_t get_record_type(const size_t tag) {return static_cast<record_t>(tag >> 9);}
```
adamnovak, line 965: I think all these are going to need to be `inline` for linking.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -822,176 +999,117 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
-#ifdef debug_distance_indexing
-            record_t type = get_record_type();
-            assert(type >= 1 && type <= 17 );
-#endif
-        }
+        SnarlTreeRecord (size_t pointer, const bdsg::yomo::UniqueMappedPointer<bdsg::MappedIntVector>* tree_records);
+        SnarlTreeRecord (const net_handle_t& net, const bdsg::yomo::UniqueMappedPointer<bdsg::MappedIntVector>* tree_records);
 
-        virtual size_t get_offset() { return record_offset; }
+        size_t get_offset() { return record_offset; }
```
adamnovak, line 1005: These also probably need to be `inline` if the bodies are here.

adamnovak: And this one can be `const`.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -822,176 +999,117 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
-            records = tree_records;
-#ifdef debug_distance_indexing
-            record_t type = get_record_type();
-            assert(type >= 1 && type <= 17 );
-#endif
-        }
+        SnarlTreeRecord (size_t pointer, const bdsg::yomo::UniqueMappedPointer<bdsg::MappedIntVector>* tree_records);
+        SnarlTreeRecord (const net_handle_t& net, const bdsg::yomo::UniqueMappedPointer<bdsg::MappedIntVector>* tree_records);
 
-        virtual size_t get_offset() { return record_offset; }
```
adamnovak, line 843: Having these not be virtual is good, I think. You probably want like one field in this struct that holds the offset, and one for the pointer, so it's cheap to construct and fast to use as an accessor.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -822,176 +999,117 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
-        //TODO: These are redeclared so that I don't need to pass the SnarlTreeRecord the actual distance index
-        //Get the offset into snarl_tree_records for a node record
-        virtual size_t get_offset_from_id (const handlegraph::nid_t id) const;
     };
 
     //Record interpreter that has a non-const reference to snarl_tree_records, so it can
     //also add things
     struct SnarlTreeRecordConstructor {
 
-        //TODO: This might be bad but this has the same members as SnarlTreeRecord
```
adamnovak, line 913: I think the next line doesn't make sense without this one.

You do sometimes have to duplicate code to deal with const-ness. I don't think you can inherit from something and make a field it had that was const not const anymore. Though you could keep it const and do a bunch of const casting.

You could try templating things on the pointer type. So you have:

```

template<typename P>
struct BaseSnarlTreeRecord {
    P records;
    size_t offset;
    ...
};

using SnarlTreeRecord = BaseSnarlTreeRecord<const bdsg::yomo::UniqueMappedPointer<bdsg::MappedIntVector>*>;
struct MutableSnarlTreeRecord: public BaseSnarlTreeRecord<bdsg::yomo::UniqueMappedPointer<bdsg::MappedIntVector>*> {
...
};
```

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -822,176 +999,117 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
-        virtual size_t get_connected_component_count() const {return (*records)->at(record_offset+COMPONENT_COUNT_OFFSET);}
-        virtual size_t get_node_count() const {return (*records)->at(record_offset+NODE_COUNT_OFFSET);}
-        virtual size_t get_max_tree_depth() const {return (*records)->at(record_offset+MAX_TREE_DEPTH_OFFSET);}
-        virtual size_t get_min_node_id() const {return (*records)->at(record_offset+MIN_NODE_ID_OFFSET);}
-        virtual SnarlTreeRecord get_component_record(size_t component_number) const {return SnarlTreeRecord((*records)->at(record_offset+2+component_number), records);}
-        virtual bool for_each_child(const std::function<bool(const handlegraph::net_handle_t&)>& iteratee) const;
+        RootRecord (size_t pointer, const bdsg::yomo::UniqueMappedPointer<bdsg::MappedIntVector>* tree_records);
+        RootRecord (net_handle_t net, const bdsg::yomo::UniqueMappedPointer<bdsg::MappedIntVector>* tree_records);
+
+        size_t get_connected_component_count() const {return (*records)->at(record_offset+COMPONENT_COUNT_OFFSET);}
```
adamnovak, line 1107: These might also need to be inline.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -805,14 +978,18 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+/* Define a struct for interpreting each type of snarl tree node record (For node, snarl, chain)
+ *
+ * This is meant to be a layer in between snarl_tree_records and the public interface
+ * Each net_handle_t has a pointer to the start of a record in snarl_tree_records. 
+ * SnarlTreeRecord keeps the pointer and interprets the values stored in the record 
+ *
+ * SnarlTreeRecordConstructor does the same thing but for writing values to the index
+ *
+ */
     struct SnarlTreeRecord {
```
adamnovak, line 993: These are cheap abstractions, but I don't think they're quite zero-cost, since you need to spend time arranging the pointers and offsets into the right memory layout somewhere to form the object to call the methods, and following `this` pointers and possibly calling non-inlined functions. Just keeping everything in arguments to static `inline` functions to do the accesses might be slightly faster.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -822,176 +999,117 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+        handlegraph::nid_t get_end_orientation() const;
 
-        //TODO: These are redeclared so that I don't need to pass the SnarlTreeRecord the actual distance index
-        //Get the offset into snarl_tree_records for a node record
-        virtual size_t get_offset_from_id (const handlegraph::nid_t id) const;
     };
 
     //Record interpreter that has a non-const reference to snarl_tree_records, so it can
     //also add things
     struct SnarlTreeRecordConstructor {
```
adamnovak, line 1067: I wouldn't call these "Constructor" since that's already a thing. Maybe `MutableSnarlTreeRecord` or `SnarlTreeRecordWriter` wight be better?

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -8,11 +8,45 @@
 #include <handlegraph/algorithms/dijkstra.hpp>
 #include <handlegraph/util.hpp>
 #include <handlegraph/trivially_serializable.hpp>
+#include <handlegraph/algorithms/dijkstra.hpp>
 #include <bdsg/internal/mapped_structs.hpp>
 #include <string>
+#include <numeric>
 #include <arpa/inet.h>
+#include <jansson.h>
```
adamnovak, line 16: I don't think you actually use any Jansson types in this header. In that case, it would be better to put the include in the CPP file.

Otherwise, anyone who includes the distance index header also has to read the Jansson header, and the compiler will waste a bunch of time reading the Jansson header when it isn't going to be needed to build that `.o` file.

adamnovak (commented):
This is really only a partial review, since I haven't gotten to the main CPP. And since the PR doesn't actually bring in the whole index at once so I can only comment the changed lines.

What I've gotten through looks OK, although I think there might need to be more `inline` to appease the linker.

I still need to go through the full CPP and form some opinions on it and whether I can see anything that is confusing or might be slow.

adamnovak:
Why are the record shape definition constants part of the index itself, and not part of the little accessor classes?

bdsg/docs/rst/api.rst
```diff
@@ -117,7 +121,18 @@ Orthogonal to the mutability and paths hierarchy, there is a :class:`bdsg.handle
    :show-inheritance:
    :members:
    :inherited-members:
-   
+ 
+~~~~~~~~~~~~~~~~~~
+SnarlDecomposition
+~~~~~~~~~~~~~~~~~~
+
+A "snarl decomposition" describes the decomposition of the graph into nested substructures known as snarls and chains. The :class:`bdsg.handlegraph.SnarlDecomposition` interface defines methods for traversing the snarl decomposition of a graph using :class:`bdsg.handelgraph.net_handle_t`.
```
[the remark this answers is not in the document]
xchang1: I can write one. I'm not sure where the right place for it is though, either in the vg wiki or the bdsg tutorial?

bdsg/docs/conf.py
```diff
@@ -57,7 +57,7 @@
 
 # -- Breathe Configuration ---------------------------------------------------
 
-breathe_projects = { "bdsg": "../build/doxygen/xml" }
+breathe_projects = { "bdsg": "../../build/doxygen/xml" }
```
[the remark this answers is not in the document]
xchang1: I think so, all the errors went away at least

xchang1 (commented):
@adamnovak I got it to find jansson, can you double check that this is ok, please?
I mean the changes to CMakeLists in commit [030444c](https://github.com/vgteam/libbdsg/pull/165/commits/030444c4460b7098a658c94630c9dee254b38a73)

Issue #168, opened 2022-11-04
Speed up finding reference paths on a handle: Make PathPositionOverlay also implement its own for_each_step_on_handle_impl?


adamnovak:
For surjection, we work with path position overlays for a restricted subset of paths in the backing graph. But searching for all the steps on a handle can involve thinking about all the paths *actually* in the backing graph, and checking each of them. We would like to be able to just get the reference path steps on the handle efficiently.

One way to do this might be to re-implement `for_each_step_on_handle_impl` based on the information in the overlay that knows about the subsetting of the paths.

https://github.com/vgteam/libbdsg/blob/06ab14bbc7fcfd40745885cd85c301e4bf1cf0cd/bdsg/src/path_position_overlays.cpp#L149-L152

adamnovak:
The `MemoizingGraph` in vg isn't much help here because we want to answer these queries in random order across the graph, and it only memoizes the most recently queried handles' path steps.

PR #169, opened 2022-11-09
Add overlay that accelerates path lookup on nodes

adamnovak:
This adds a `PackedReferencePathOverlay` which, in addition to what the `PackedPositionOverlay` does, speeds up `for_each_step_on_handle()` by keeping a mapping from node ID to step rank in each path's assigned `PackedPositionOverlay` index object.

To actually do the query, we still have to poll all the indexes which got generated in parallel; we're not creating one big mapping from node ID to relevant index. I'd *like* to avoid having to do any single-threaded work like that over the whole graph, because that sort of thing was too slow and we had to move to index the different paths in parallel.

I'm doing some profiling to see if this is good enough or if I really do want to coalesce the lookup tables or add another global lookup table.


adamnovak:
This will fix #168

glennhickey:
Awesome!  Once it's merged, I'll test it on a more substantial graph.

PR #182, opened 2023-02-09
Allow preloading the distance index from disk

adamnovak:
This adds machinery to the distance index and the memory-mapping chains to let you blockingly bring the whole index into memory, or else hint that it should be all brought into memory. This can be a lot faster than waiting for the kernel to fault all the pages as they are accessed, because it gets to use sequential and not random disk reads.

### DIFF
--- a/bdsg/include/bdsg/internal/mapped_structs.hpp
+++ b/bdsg/include/bdsg/internal/mapped_structs.hpp
@@ -311,6 +311,15 @@ public:
     static void* allocate_from_same_chain(void* here, size_t bytes);
     
     /**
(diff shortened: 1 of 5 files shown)


adamnovak:
I don't know what's up with the EINVALs; I was getting a successful madvise for things under a page, but failures for like a whole page or anything over it, in the tests, and then failures for loading from a file in Giraffe. When it fails it seems like it fails for the whole range and for any single page in the range, too. Maybe mapping from a file with MIO is applying one of the mystery flags in the kernel that are incompatible with MADV_POPULATE_READ?

PR #190, opened 2023-09-09
Update net_handle_t bindings to include equality and hash

xchang1:
I wanted `net_handle_t`'s to be used in sets in python so I needed the bindings to include equality and hashing.

@adamnovak or maybe @ctcisar, can I do this? It looks like the binding code is automatically generated so I'm worried it'll get overwritten, but this was the only way I could figure out to get it to do what I wanted


adamnovak:
I don't *think* this will stick around. I think we would need a member `operator==` for BInder to bind it to Python's equality check: https://cppbinder.readthedocs.io/en/latest/limitations.html#external-operators

If we want to customize the bindings for a type, we can use an `+add_on_binder` directive in the Binder config file, break this code out into its own function that takes I think `cl` as an argument, and then get Binder to generate the code to call that function: https://cppbinder.readthedocs.io/en/latest/config.html?highlight=add_on_binder#config-file-directives

adamnovak (approved):
The only change is updating libhandlegraph and building the bindings.

But we probably need a PR on the libhandlegraph side.

PR #199, opened 2024-06-18
A simpler overlay to speed up reference path queries

jeizenga:
Includes a new overlay, `ReferencePathOverlay`, that has similar functionality to the `PackedReferencePathOverlay` except with simpler backing data structures that should be faster to access. The construction algorithm is pretty parallel, so hopefully it won't add too much to start-up time. 

I've stolen the name `ReferencePathOverlayHelper` to refer to the `ReferencePathOverlay`. The older overlay helper is now named `PackedReferencePathOverlayHelper`.

@adamnovak This does PR does not include any updates to the Python bindings. I'm not sure if I have easy access to a computational environment where I can run the binder code. Is this mergeable without that?


adamnovak:
I'm not sure we ever bound the overlay helpers, so there probably isn't a breaking change vs. the existing generated binding code we have checked in.

PR #213, opened 2025-06-13
Paths bug

xchang1:
I was having a problem where I used an overlay to get a `PathPositionHandleGraph`, and the graph would call `PathMetadata::for_each_path_matching_impl` instead of `GBWTGraph::for_each_path_matching_impl` (for example). 

I tried to fix it  [here](https://github.com/vgteam/libhandlegraph/pull/101) I think this does it properly by having the PackedPositionOverlay and PathPositionOverlay implement their own version of `for_each_path_matching_impl` that calls the graph's version of the method, instead of defaulting to `PathMetadata`'s versions.


0920911554 (approved):

xchang1:
@adamnovak it looks like a bot approved my PR, idk if this is a problem

PR #214, opened 2025-06-18
Index path positions

xchang1:
This makes PackedPositionOverlay index all paths using `for_each_path_matching`, instead of only reference and generic paths using `for_each_path_handle`.
@adamnovak I'm not sure if this is what it's supposed to do but it's what I want it to do.


adamnovak:
@xchang1 I merged this, but it looks like the merge commit is an empty commit with no changes in its diff. Can you verify that the `master` branch now actually has the behavior you want?

PR #215, opened 2025-06-23
Make PackedPositionOverlay index all paths

xchang1:
This does what [#214](https://github.com/vgteam/libbdsg/pull/214) wanted to but actually does it this time


adamnovak:
This is going to include things like `_alt` paths from VCF construction, right?

adamnovak:
The way we do this in `ReferencePathOverlay` is to index only the presented (i.e. non-haplotype and I think non-alt?) paths, but provide a way to hint the names of other paths you also want indexed.

PR #218, opened 2025-07-16
Check if handlegraph_objs already exists before building it

xchang1:
I was having trouble building a Cmake project using `libbdsg` and `libvgio` because both depended on `libhandlegraph`. 

```
CMake Error at deps/libvgio/deps/libhandlegraph/CMakeLists.txt:37 (add_library):
  add_library cannot create target "handlegraph_objs" because another target
  with the same name already exists.  The existing target is created in
  source directory
  "/home/xhchang/GitHub/STOAT/deps/libbdsg/bdsg/deps/libhandlegraph".  See
  documentation for policy CMP0002 for more details.


CMake Error at deps/libvgio/deps/libhandlegraph/CMakeLists.txt:116 (add_library):
  add_library cannot create target "handlegraph_shared" because another
  target with the same name already exists.  The existing target is a shared
  library created in source directory
  "/home/xhchang/GitHub/STOAT/deps/libbdsg/bdsg/deps/libhandlegraph".  See
  documentation for policy CMP0002 for more details.


CMake Error at deps/libvgio/deps/libhandlegraph/CMakeLists.txt:119 (add_library):
  add_library cannot create target "handlegraph_static" because another
  target with the same name already exists.  The existing target is a static
  library created in source directory
  "/home/xhchang/GitHub/STOAT/deps/libbdsg/bdsg/deps/libhandlegraph".  See
  documentation for policy CMP0002 for more details.
```

I fixed this by checking if `handlegraph_objs` already exists before re-building it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,10 +175,16 @@ target_link_libraries(dynamic INTERFACE tsl::hopscotch_map)
 
 # libhandlegraph (full build using its cmake config)
 # Produces handlegraph_shared and handlegraph_static targets (as well as handlegraph_objs)
-if (NOT TARGET handlegraph_objs)
-    # If handlegraph has not already been built by something else
+find_package(libhandlegraph)
+if (${libhandlegraph_FOUND})
+    message("Using installed libhandlegraph")
+elseif (NOT TARGET handlegraph_objs)
+    message("Using bundled libhandlegraph")
     add_subdirectory("${bdsg_DIR}/deps/libhandlegraph")
+else ()
+    message("Using libhandlegraph built by someone else")
 endif()
+
 
 # BBHash perfect hasher (header only)
 # Does not ship its own install step or define a target, so we make our own target
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,7 +175,10 @@ target_link_libraries(dynamic INTERFACE tsl::hopscotch_map)
 
 # libhandlegraph (full build using its cmake config)
 # Produces handlegraph_shared and handlegraph_static targets (as well as handlegraph_objs)
-add_subdirectory("${bdsg_DIR}/deps/libhandlegraph")
+if (NOT TARGET handlegraph_objs)
+    # If handlegraph has not already been built by something else
+    add_subdirectory("${bdsg_DIR}/deps/libhandlegraph")
+endif()
 
 # BBHash perfect hasher (header only)
 # Does not ship its own install step or define a target, so we make our own target


xchang1:
On second thought, maybe this isn't a good idea if the `handlegraph_objs` that was already built isn't the right version for this version of `bdsg`

xchang1:
It works for me if I build the version of `libhandlegraph` that I want, then make both `bdsg` and `vgio` use that version instead of their own. But it still depends on both of them being able to work with the version of libhandlegraph that I use

adamnovak:
There's no way to have multiple versions of `libhandlegraph` actually linked into the same C++ project, that I know of. So if you link `libbdsg` and `libvgio` both in, they need to both be compatible with the one final `libhandlegraph`. So in that case it should be fine to have one copy in the CMake setup turn itself off and let the other copy win.

CMakeLists.txt
```diff
@@ -175,7 +175,16 @@ target_link_libraries(dynamic INTERFACE tsl::hopscotch_map)
 
 # libhandlegraph (full build using its cmake config)
 # Produces handlegraph_shared and handlegraph_static targets (as well as handlegraph_objs)
-add_subdirectory("${bdsg_DIR}/deps/libhandlegraph")
+find_package(libhandlegraph)
+if (${libhandlegraph_FOUND})
+    message("Using installed libhandlegraph")
```
adamnovak, line 180: Preferring the system-installed `libhandlegraph` *does* seem like a potential source of compatibility problems, unless we start versioning it. The vg build would use this path, I guess, since it provides its own libhandlegraph and injects it into the search paths first, but someone with a working `libbdsg` CMake project might have their project start failing to build if they install a system `libhandlegraph` of the wrong version.

We probably want some CMake flags you can pass to always use the bundled libhandlegraph, or a version requirement here.

adamnovak (changes requested):

adamnovak (approved):

CMakeLists.txt
xchang1: I think I actually want it to default to using the version of `libhandlegraph` that got made by my cmake that calls libbdsg's cmake. So I made a new [pr](https://github.com/vgteam/libbdsg/pull/219)

Is this what you meant by adding a CMake flag? It's not very pretty with the repetitive conditionals, but now it will use the installed version if the flag is set and it exists, and otherwise it will default to looking for an existing version, and if that doesn't exist it will fall back on using the bundled version.

xchang1: One of my colleagues also mentioned that there might be a problem with the linker, that it might default to looking for the `.so` files in `/usr/local` or somewhere else, instead of using the one that cmake built.

adamnovak: If you're using the bundled libhandlegraph, I think you need the CMake install step to know it also has to install that. I *think* that will work by default when you create library targets.

And you would indeed have to rely on wherever it installs to being a higher priority in the library search path for the final executable or downstream library user. But if you have a binary that's picking up non-system-installed .so files at all, it *probably* is searching those non-system directories first.

xchang1: I think it's installing the bundled libhandlegraph? I see `libhandlegraph.so` and `libhandlegraph.a` in `build/deps/libhandlegraph`, that is what's supposed to happen when it installs it, right? 

I also tried installing a very old version of libhandlegraph (from 2019) that I think shouldn't work with the current versions of libbdsg and libvgio. If I use the new flag `-DUSE_INSTALLED_LIBHANDLEGRAPH=ON`, it compiles but when I try to run it I get an error `./stoat: symbol lookup error: /home/xhchang/GitHub/STOAT/build/libgbwtgraph.so: undefined symbol: _ZNK11handlegraph11HandleGraph14get_edge_countEv`. When I don't pass it the flag to use the installed libhandlegraph everything works fine. So it looks like it's linking the correct version?

xchang1: Also, rereading your previous comment, I think I did the flag backwards. Should it be defaulting to using the installed `libhandlegraph` and take a flag to use the bundled one instead?

PR #237, opened 2026-02-06
add all-paths toggle to packed reference overlay

glennhickey:
Same as #236 but for the PackedReferenceOverlay (which vg call uses).


adamnovak:
I think we have the same basic problem as this with `vg find`, which uses a plain `PathPositionOverlay`, which still thinks "hidden" paths can exist and is indexing everything instead of just the paths `vg find` is actually trying to look things up on.

PR #238, opened 2026-02-13
Save orientations of tips in tippy_child_ranks

faithokamoto:
I needed to know which end of the child tip was the tip end (i.e. has no edges) so I added a field to `tippy_child_ranks`. Barely affects anything since those data weren't really used.


faithokamoto:
Oh yeah, and I'm bumping the version number since this is a vg giraffe bugfix.

PR #239, opened 2026-04-07
Hublabel

electricEpilith:
libbdsg changes to go with merging the hublabel branch of vg


bdsg/include/bdsg/internal/indexing_iterator.hpp
```diff
@@ -0,0 +1,208 @@
...
+    /// Subtraction of offset
+    IndexingIterator operator-(difference_type offset) const;
+
+    /// Subtraction of two iterators
+    difference_type operator-(const IndexingIterator& other) const;
+
+    /// Indexing into iterator. Even though we type this as reference, remember
+    /// that we don't actually implement writing to our "references" and just
+    /// use the value type.
+    /// Result is undefined if itrators are to different collecitons.
```
adamnovak:
```suggestion
    /// Result is undefined if iterators are to different collections.
```

bdsg/include/bdsg/internal/indexing_iterator.hpp
```diff
@@ -0,0 +1,208 @@
...
+    reference operator[](difference_type offset) const;
+
+    // Comaprable iterator methods (TODO: Is there an STL concept name for this?)
+    
+    /// Determine if this iterator is strictly before another.
+    /// Result is undefined if iterators are to different collecitons.
```
adamnovak:
```suggestion
    /// Result is undefined if iterators are to different collections.
```

bdsg/include/bdsg/internal/indexing_iterator.hpp
```diff
@@ -0,0 +1,208 @@
...
+    bool operator<(const IndexingIterator& other) const;
+
+    /// Determine if this iterator is before or at another.
+    /// Result is undefined if iterators are to different collecitons.
```
adamnovak:
```suggestion
    /// Result is undefined if iterators are to different collections.
```

bdsg/include/bdsg/internal/indexing_iterator.hpp
```diff
@@ -0,0 +1,208 @@
...
+    bool operator<=(const IndexingIterator& other) const;
+
+    /// Determine if this iterator is strictly after another.
+    /// Result is undefined if iterators are to different collecitons.
```
adamnovak:
```suggestion
    /// Result is undefined if iterators are to different collections.
```

bdsg/include/bdsg/internal/indexing_iterator.hpp
```diff
@@ -0,0 +1,208 @@
...
+    bool operator>(const IndexingIterator& other) const;
+
+    /// Determine if this iterator is at or after another.
+    /// Result is undefined if itrators are to different collecitons.
```
adamnovak:
```suggestion
    /// Result is undefined if iterators are to different collections.
```

bdsg/include/bdsg/internal/indexing_iterator.hpp
```diff
@@ -0,0 +1,208 @@
+#ifndef BDSG_INDEXING_ITERATOR_HPP_INCLUDED 
+#define BDSG_INDEXING_ITERATOR_HPP_INCLUDED
+
+namespace bdsg {
+/**
+ * A forward iterator for anything vector-like (PackedVector, PagedVector,
```
adamnovak, line 8: Right now it's not *anything* vector-like; it demands a `get()` method (which we then turn around and add to at least one of the vector types). Why can't it just use `operator[]`? We could even maybe make `reference` be `decltype(std::declval<VectorLike>()[0])`.

bdsg/include/bdsg/internal/mapped_structs.hpp
```diff
@@ -1860,6 +1879,7 @@ CompatIntVector<Alloc>::ConstProxy::operator uint64_t () const {
 template<typename Alloc>
 auto CompatIntVector<Alloc>::at(size_t index) -> Proxy {
     if (index > size()) {
+        assert(false);
```
adamnovak:
```suggestion
```

We probably don't want both the assert and the throw.

bdsg/include/bdsg/internal/mapped_structs.hpp
```diff
@@ -1869,6 +1889,7 @@ auto CompatIntVector<Alloc>::at(size_t index) -> Proxy {
 template<typename Alloc>
 auto CompatIntVector<Alloc>::at(size_t index) const -> ConstProxy {
     if (index > size()) {
+        assert(false);
```
adamnovak:
```suggestion
```

We probably don't want both the assert and the throw.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,281 @@
+/*
+file for quickly playing around with stuff
```
adamnovak: This needs a better description; we can't mainline a file for quickly playing around with stuff.

The file might also need a more descriptive file name.

What determines if a piece belongs here or in `hublabel.hpp` or `landmark.hpp`?

Also, if there's not a class that represents the hub labeling module, this is where you would document the whole module: the concepts involved, citations to the literature, what functions the user of the module is expected to call in sequence to actually use it, and so on.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,281 @@
+/*
```
adamnovak, line 1: Include guards are missing for this header.

bdsg/include/bdsg/hublabel.hpp
```diff
@@ -0,0 +1,65 @@
+/*
```
adamnovak: Include guards are missing for this header.

bdsg/include/bdsg/landmark.hpp
```diff
@@ -0,0 +1,270 @@
...
+#define LANDMARK_HPP
+
+#include <span>
+#include <ranges>
+#include <boost/heap/priority_queue.hpp>
+#include <boost/multi_array.hpp>
+#include <bdsg/hash_graph.hpp>
+#include <sdsl/vectors.hpp>
+//#include <catch2/catch_session.hpp>
+//#include <catch2/catch_test_macros.hpp>
```
adamnovak: This commented-out code should be removed.
```suggestion
```

bdsg/include/bdsg/landmark.hpp
```diff
@@ -0,0 +1,270 @@
...
+
+using namespace std;
+namespace bdsg {  
+//inf implementation is largest possible int
+#define INF_INT numeric_limits<int>::max()
+#define DIST_NBITS 32
+#define DIST_UINT uint32_t
+#define ARR2D_OFFSET 1
```
adamnovak: These should probably be replaced with `constexpr` variables and `using`s so they can be namespaced under `bdsg`.

bdsg/include/bdsg/landmark.hpp
```diff
@@ -0,0 +1,270 @@
...
+    if (val > (OtherInt) INF_INT) {
+        throw std::overflow_error("Cannot store excessively wide value " + std::to_string(val) + " in " + std::to_string(DIST_NBITS) + " bits for hub labeling");
+    }
+    return (DIST_UINT) val;
+}
+
+/// Sum two distances, propagating infinities.
+/// Does not check for overlfow.
+/// TODO: We're not really sure if our distances are ints or uints and we freely mix them when we shouldn't.
+int addInt(int a, int b);
```
adamnovak: We should solve this problem and make sure all distances are consistently DIST_UINT and never `int`.

bdsg/include/bdsg/landmark.hpp
```diff
@@ -0,0 +1,270 @@
...
+
+typedef int NodeId;
+typedef int NodesideId;
+typedef enum EnterDir {OTHER_NODESIDE=0,OTHER_NODE=1} EnterDir;
+typedef boost::multi_array<int, 2> Array2D;
+typedef unordered_map<NodesideId, int> Ball;
+class SdslArray2D {
```
adamnovak: This should have a doc comment explaining what it is.

bdsg/include/bdsg/landmark.hpp
```diff
@@ -0,0 +1,270 @@
...
+string nodeside_string(NodesideId nodeside);
+
+
+/// landmark selection
+
+/*
+returns ids of nodesides with largest degree
+*/
+//helper
+int nodeside_degree(NodesideId ns_id, HashGraph& g);
```
adamnovak: I think here `landmark selection` ends up being the doc comment for the function.

bdsg/include/bdsg/landmark.hpp
```diff
@@ -0,0 +1,270 @@
...
+
+void get_closest_lm_ind(SdslArray2D& dist_table, vector<int>& ind_vec, vector<int>& min_dist_vec, int nodeside_count);
+
+/*
+based off https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/msr-tr-2009-84.pdf [Chen et al. (2009)]
+*/
+Ball get_ball_contents(NodesideId ns, int dist_limit, HashGraph& g);
+
+
+vector<Ball> find_balls(vector<int>& closest_lm_dist_vec, HashGraph& g, int min_ball_size = 0);
```
adamnovak: I think the whole ball system is not actually used.

I think in general there's like 3 versions of hub labeling in here but only `make_boost_graph()`, `make_contraction_hierarchy()`, `create_labels()`, `pack_labels()`, and `hhl_query()` are actually used.

All the functions that they don't call, and types not involved with the actually-used call tree, should be removed.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -621,40 +624,165 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+        if (is_multicomponent) {
+            if (!has_distances) {
+                throw runtime_error("error: multicomponent chain must have distances");
+            }
+            return MULTICOMPONENT_CHAIN;
+        } else {
+            return has_distances ? DISTANCED_CHAIN : CHAIN;
+        }
+    }
+
```
adamnovak, line 844: All of these probably ought to have doc comments.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -822,12 +950,12 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
-    // While the version number is 4, store it in a bit masked way
+    // We store the version number in a bit masked way
     // to avoid getting confused with old indexes without version numbers
     // that start with component count
-    const static size_t CURRENT_VERSION_NUMBER = 4;
-    // A verion to allow though but warn about
-    const static size_t WARN_VERSION_NUMBER = 3;
+    const static size_t CURRENT_VERSION_NUMBER = 5;
+    // A version to allow though but warn about
+    const static size_t WARN_VERSION_NUMBER = 4;
```
adamnovak: We might want to reject the current v4 indexes, because we won't understand their oversized snarls.

Maybe we can just reject them if they have oversized snarls? I don't know if that field is available.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1268,7 +1430,10 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
 
         //How big is the entire snarl record?
         static size_t distance_vector_size(record_t type, size_t node_count);
-        static size_t record_size (record_t type, size_t node_count) ;
+        //vec_size parameter only needed for oversized snarls
+        //represents size of hub labeling-related data
+        //the value needed should be the first entry after the fixed-size record data
+        static size_t record_size (record_t type, size_t node_count, size_t vec_size) ;
```
adamnovak: This could become part of a doc comment.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1302,9 +1467,33 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
         SnarlRecordWriter(bdsg::yomo::UniqueMappedPointer<bdsg::MappedIntVector>* records, size_t pointer);
 
+        /**
+         * Set size of hub labels flat vector (hub_labels).
+         * Only used for oversized snarls.
+         *
+         * May only be called once.
+         *
+         * TODO: Make separate SnarlRecordWriter for oversized snarls?
+         *
```
adamnovak: I think we decided against this.
```suggestion
```

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1642,19 +1858,24 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
 
             bool reversed_in_parent;
             bool start_node_rev;
+            // The end node is reverse if it points back into the snarl, not along it.
```
adamnovak, line 1917: This could become part of a doc comment.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1642,19 +1858,24 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
             bool is_simple;
+            /// Set to true if the snarl is regular (see SnarlDistanceIndex::is_regular_snarl()).
+            /// If is_simple is true, this must also be set to true when filling in the TemporarySnarlRecord.
+            bool is_regular = false;
             bool is_tip = false;
             bool is_root_snarl = false;
             bool include_distances = true;
-            //All children, nodes and chains, in arbitrary order
-            vector<pair<temp_record_t, size_t>> children; 
+            vector<temp_record_ref_t> children; //All children, nodes and chains, in arbitrary order
```
adamnovak, line 1928: This could become a doc comment.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1642,19 +1858,24 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
             //The ranks & orientations of children that are tips
-            unordered_map<size_t, bool> tippy_child_ranks; 
+            unordered_map<size_t, bool> tippy_child_ranks;
             //vector<tuple<pair<size_t, bool>, pair<size_t, bool>, size_t>> distances;
             unordered_map<pair<pair<size_t, bool>, pair<size_t, bool>>, size_t> distances;
-                     
+            //linearized hub labels (if not empty, this is an oversized snarl)
```
adamnovak, line 1933:
```suggestion
            /// Linearized hub labels (if not empty, this is an oversized snarl)
```

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1302,9 +1467,33 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+         */
+        void set_vec_size(size_t vec_size);
+        
+        /**
+         * Set an entry in the vector holding the hub label data.
+         * Only used for oversized snarls.
+         *
+         * set_vec_size() must be called first.
+         */
+        void set_vec_entry(size_t index, size_t value);
```
adamnovak: We probably want to call the flattened hub label data something other than `vec`, because we also have the whole vector that the distance index lives in in play. We could say things like `hhl_size`, or a more verbose `hub_label_size`.

adamnovak: These would also change name if we rename `vec`.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1683,12 +1904,91 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+            }
+            if (ref.second < min_node_id) {
+                throw std::out_of_range("Trying to look up node " + std::to_string(ref.second) + " but temporary index starts at node " + std::to_string(min_node_id));
+            }
+            if (ref.second >= temp_node_records.size() + min_node_id) {
+                throw std::out_of_range("Trying to look up node " + std::to_string(ref.second) + " but temporary index only goes up until node " + std::to_string(temp_node_records.size() + min_node_id));
+            }
+            // Nodes use a node ID in the ref, not an index.
+            return temp_node_records[ref.second - min_node_id];
+        }
```
adamnovak, line 2047: These could use docstrings.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
+/*
+file for quickly playing around with stuff
```
adamnovak: This needs a better idea of what the file is.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+  g.create_edge(h[2], g.flip(h[7])); g.create_edge(h[3], h[5]);
+  g.create_edge(h[4], h[6]); g.create_edge(h[4], h[7]);
+  g.create_edge(h[5], h[4]); g.create_edge(h[5], h[6]);
+  g.create_edge(h[6], h[8]); g.create_edge(h[7], h[8]);
+  g.create_edge(h[7], h[9]); g.create_edge(h[7], h[10]);
+  g.create_edge(g.flip(h[7]), g.flip(h[10])); 
+  g.create_edge(h[9], g.flip(h[10])); g.create_edge(h[11], h[12]);
+  g.create_edge(h[12], g.flip(h[1]));
+  return g;
+}
```
adamnovak: Test code like this should be with the other libbdsg test code.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,281 @@
...
+  // This is left uninitialized until make_contraction_hierarchy() is run.
+  NODE_UINT new_id;
+} NodeProp;
+
+typedef struct EdgeProp { 
+  bool contracted = false;
+  DIST_UINT weight = 0;
+  NODE_UINT arc_cover = 1;
+  bool ori = true;
+} EdgeProp;
```
adamnovak: These should have a little more documentation.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,281 @@
...
+/**
+ * Build the intermediate hub labeling computation data structure ("Boost
+ * graph") from a HashGraph.
+ *
+ * The nodes in the graph must have dense node IDs starting at 1.
+ *
+ * For later queries, orientations of nodes are assigned ranks as provided by
+ * the bgid() function.
+ */
+CHOverlay make_boost_graph(const bdsg::HashGraph& hg);
```
adamnovak: Can this have a better name than `bdsg::make_boost_graph()`? And a better name for the data structure than the "Boost graph"?

The names of the hub labeling API functions we actually use don't have much to tie them together (they don't belong to a class as static methods, or to a namespace, or share any words in their names). Talking here about a "Boost graph" is telling the user about the implementation detail that should be hidden (the fact that Boost is a dependency) and not about the fact that this is hub-labeling-related or what the graph that's implemented using Boost actually represents.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,281 @@
...
+ *
+ * For later queries, orientations of children or the snarl boundary nodes are assigned query ranks based on their snarl distance index rank.
+ *
+ * The snarl distance index ranks are 0 and 1 for the start and end nodes of the snarl, and the rank_in_parent field of the temporary index for each child.
+ */
+CHOverlay make_boost_graph(const bdsg::SnarlDistanceIndex::TemporaryDistanceIndex& temp_index, const SnarlDistanceIndex::temp_record_ref_t& snarl_index, const SnarlDistanceIndex::TemporaryDistanceIndex::TemporarySnarlRecord& temp_snarl_record, const vector<pair<SnarlDistanceIndex::temp_record_t, size_t>>& all_children, const HandleGraph* graph);
+
+int edge_diff(ContractedGraph::vertex_descriptor nid, ContractedGraph& ch, CHOverlay& ov, vector<DIST_UINT>& node_dists, int hop_limit);
+
+void contract(CHOverlay::vertex_descriptor nid, ContractedGraph& ch, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<bool>& shouldnt_contract, int hop_limit);
```
adamnovak: These should maybe be `static` since they are file-scope helpers. It would be good if they had some doc comments.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,281 @@
...
+  } else {
+    //forwards label
+    auto first_fwd_bound_itr = next(start_itr, 1); 
+    auto last_fwd_bound_itr = next(start_itr, 1+node_count);
+    auto jump_to_dist = (*last_fwd_bound_itr) - *first_fwd_bound_itr;
+    return next(hub_itr, jump_to_dist); 
+  }
+} 
+
+DIST_UINT binary_intersection_ch(vector<HubRecord>& v1, vector<HubRecord>& v2);
```
adamnovak: This could use a doc comment.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,281 @@
...
+  return dist; 
+}      
+
+void down_dijk(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_rev);
+
+void down_dijk_rev(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_rev);
+
+void test_dijk(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_rev);
+
+void test_dijk_rev(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_rev);
```
adamnovak: Test code should move to the test files.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,281 @@
...
+
+void create_labels(vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_rev, CHOverlay& ov);
```
adamnovak: This should have a doc comment.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,281 @@
...
+  
+  DIST_UINT dist = binary_intersection_ch(start_itr, start_index_1, start_index_2);
+
+
+  return dist; 
+}      
+
+void down_dijk(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_rev);
+
+void down_dijk_rev(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_rev);
```
adamnovak: What's the difference between these?

Also, it would be good to organize this file so the things that are part of the external interface are clearly separated from the ones (like these) that aren't. A class with public and private methods/types could be used to formalize that.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,281 @@
...
+ * label count | start offsets (fwd) | start offsets (back) | fwd label hubs | fwd label dists | back label hubs | back label dists 
+*/
+vector<size_t> pack_labels(const vector<vector<HubRecord>>& labels, const vector<vector<HubRecord>>& labels_back); 
+
+//not necessary stuff
+void write_to_csv(CHOverlay& ov, string out_path);
+
+void write_to_gr(CHOverlay& ov, string out_path); 
+
+vector<CHOverlay::vertex_descriptor> read_node_order(string in_path);
```
adamnovak: I think these would want to be cut, probably.

bdsg/include/bdsg/hublabel.hpp
```diff
@@ -0,0 +1,65 @@
...
+DIST_UINT hl_build_intersect(HubRecsPair& l1, HubRecsPair& l2, vector<DIST_UINT>& node_lens); 
+
+DIST_UINT get_degree(NodeId node, bdsg::HashGraph& g);
+
+void pruned_dijkstra(bdsg::HashGraph& g, NodesideId start, vector<HubRecsPair>& labels, vector<DIST_UINT>& path_lengths, vector<NodesideId>& rank_to_ns, vector<NODE_UINT>& ns_to_rank, vector<DIST_UINT>& rank_node_lens,int stop_dist = INF_INT);
+
+DIST_UINT hl_query(NodesideId i, NodesideId j, vector<HubRecsPair>& labels, bdsg::HashGraph& g);
+DIST_UINT hl_query(NodesideId i, NodesideId j, vector<HubRecsPair>& labels, vector<DIST_UINT>& rank_node_lens);
+
+vector<HubRecsPair> make_labels(bdsg::HashGraph& g);
```
adamnovak: I think almost everything in this file, other than `HubRecord`, isn't actually used.

bdsg/include/bdsg/landmark.hpp
```diff
@@ -0,0 +1,270 @@
...
+*/ 
+typedef tuple<int,EnterDir,NodesideId> QueueObj; 
+bool pqcomp1(const QueueObj& o1, const QueueObj& o2);                    
+
+typedef std::priority_queue<QueueObj, vector<QueueObj>, function<bool(const QueueObj&, const QueueObj&)>> PriorityQueue;
+
+PriorityQueue initQueue(
+  int nodeside_count, 
+  NodesideId start_ns
+);
```
adamnovak: This ends up as `bdsg::initQueue()`; it probably should live in a class or namespace or something.

It also ought to be explained.

bdsg/include/bdsg/landmark.hpp
```diff
@@ -0,0 +1,270 @@
...
+
+/*
+Called when new nodeside is discovered.
+If path to the nodeside is the shortest found so far:
+  update path_lengths and add new queue entry for it
+*/
+bool discover_nodeside(NodesideId new_nodeside, EnterDir enter_direction, int new_len, vector<vector<int>>& path_lengths, PriorityQueue& nodeside_queue); 
+
+
+SdslArray2D dijkstra(bdsg::HashGraph& g, NodesideId start, NodesideId stop_ns = -1, int stop_dist = INF_INT, bool ball_ver = false);
```
adamnovak: I think this whole piece might be one unit? It should be wrapped up into a unit with an defined boundary and a good hub-labeling-related name, so we don't have `bdsg::dijkstra()` floating around the namespace along with what looks like its guts.

adamnovak: If the hub labeling system we actually use goes through `make_boost_graph()`, what are we doing doing dijkstra traversals on `HashGraph`s? And why `HashGraph`s and not any handle graph?

Is this piece not actually needed anymore?

bdsg/include/bdsg/landmark.hpp
```diff
@@ -0,0 +1,270 @@
...
+  SdslArray2D& lm_sides_to_ns;
+} OracleInfo;
+
+
+/*
+query algorithm from Chen et al. (2009)
+with modifications for our purpose
+*/
+
+int oracle_query(NodesideId source, NodesideId target, OracleInfo& oracle, HashGraph& g);
```
adamnovak: I don't think we use this.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+
+#ifdef debug_boost_graph
+  cerr << "=== make_boost_graph complete ===" << endl;
+#endif
+  return ov;
+}
+
+
+int edge_diff(ContractedGraph::vertex_descriptor nid, ContractedGraph& ch, CHOverlay& ov, vector<DIST_UINT>& node_dists, int hop_limit = 2) {
+  //using namespace boost;
```
adamnovak:
```suggestion
```

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    vector<int> to_reset;
+    //five hops limit idea from https://turing.iem.thm.de/routeplanning/hwy/contract.pdf
+    while ((!q.empty()) && (num_iter < hop_limit)) {
+      auto [cur_dist, cur_node] = q.top();
+      to_reset.push_back(cur_node);
+      if (cur_dist > stop_dist) { break; }
+      q.pop();
+      
+      std::tie(_, __) = out_edges(cur_node, ch);
+      std::for_each(_, __, [&](auto edge) {
```
adamnovak: If we put stuff in variables named `_` and `__`, we shouldn't actually *use* it.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+
+    for (auto n: to_reset) { node_dists[n] = INF_INT; }
+
+    while (!q.empty()) { node_dists[get<1>(q.top())] = INF_INT; q.pop(); }
+  });  
+
+  int edel = out_degree(nid, ch) + in_degree(nid, ch);
+  int ediff = eadd - edel;
+  
+  return ediff;
```
adamnovak, line 434: It would be good if this had comments sufficient to teach the algorithm to someone who hadn't read the paper (like me).

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    while (!q.empty()) { node_dists[get<1>(q.top())] = INF_INT; q.pop(); }
+  });  
+
+  int edel = out_degree(nid, ch) + in_degree(nid, ch);
+  int ediff = eadd - edel;
+  
+  return ediff;  
+}
+
+void contract(CHOverlay::vertex_descriptor nid, ContractedGraph& ch, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<bool>& shouldnt_contract, int hop_limit = 2) {
```
adamnovak: I don't like variable names like `shouldnt_contract` where we need to drop an apostrophe.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+  //update contracted neighbor counts
+  std::for_each(in_start, in_end, [&](ContractedGraph::edge_descriptor eid) {
+    auto in_node = source(eid, ch);
+    ov[in_node].contracted_neighbors += 1;
+    ov[in_node].level = max(ov[in_node].level, ov[nid].level); 
+    ch[eid].contracted = true;
+    shouldnt_contract[in_node] = true;  
+  });
+
+  //std::tie(out_start, out_end) = out_edges(nid, ch);
```
adamnovak:
```suggestion
```

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+  std::for_each(out_start, out_end, [&](ContractedGraph::edge_descriptor eid) {
+    auto out_node = target(eid, ch);
+    ov[out_node].contracted_neighbors += 1;  
+    ov[out_node].level = max(ov[out_node].level, ov[nid].level); 
+    ch[eid].contracted = true;  
+    shouldnt_contract[out_node] = true;  
+  }); 
+
+  ov[nid].contracted = true;
+}
```
adamnovak: It would be good to have comments in here explaining the algorithm.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+  vector<DIST_UINT> node_dists(num_vertices(ov), INF_INT);  
+
+  vector<int> nodes; nodes.resize(num_vertices(ov));
+  iota(nodes.begin(), nodes.end(), 0); 
+
+  for (int rnd = 0; rnd < 1; rnd++) {  
+    std::fill(skip.begin(), skip.end(), false);  
+    
+    //for (NODE_UINT i = 0; i < num_vertices(ov); i+=1) {
+    for (auto i: nodes) {
```
adamnovak: The vector here seems superfluous.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+  for (auto i = 0u; i < num_vertices(ov); i+=1) {
+    v2.emplace_back(in_degree(i,ov)*out_degree(i,ov), i);
+  }
+  sort(v2.rbegin(), v2.rend());
+  /* 
+  for (int i: {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}) {
+    auto [p, n] = v2[i];
+    cerr << n << " (" << ov[n].new_id << "): " << in_degree(n,ov) << " " << out_degree(n,ov) << endl;
+  } */
+}
```
adamnovak, line 721: There's a bunch of commented-out code in here, and only a few comments explaining how this ends up making a contraction hierarchy.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    //preparing for next pop
+    pop_heap(queue_objs.begin(), queue_objs.end(), greater<tuple<int, CHOverlay::vertex_descriptor>>());
+    
+    int hop_limit = 1000;//get_hop_limit(ov);
+    int edif = edge_diff(node, contracted_g, ov, node_dists, hop_limit);
+   
+    int new_pri = ((2*edif)+ (1*ov[node].contracted_neighbors)) + (5*(ov[node].level+1)) + ov[node].arc_cover;
+    
+    if (new_pri > get<0>(queue_objs.back())) {
+      // After recomputing priority, the priority is actually greater than the nex-tlowest-priority entry.
```
adamnovak:
```suggestion
      // After recomputing priority, the priority is actually greater than the next-lowest-priority entry.
```

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+ 
+  while (queue_objs.size() > 2) { 
+    auto [pri, node] = queue_objs.back(); queue_objs.pop_back();
+    //preparing for next pop
+    pop_heap(queue_objs.begin(), queue_objs.end(), greater<tuple<int, CHOverlay::vertex_descriptor>>());
+    
+    int hop_limit = 1000;//get_hop_limit(ov);
+    int edif = edge_diff(node, contracted_g, ov, node_dists, hop_limit);
+   
+    int new_pri = ((2*edif)+ (1*ov[node].contracted_neighbors)) + (5*(ov[node].level+1)) + ov[node].arc_cover;
```
adamnovak: Some of these variables seem excessively abbreviated. Do we need to go all the way down from `edge_difference` to `edif`? Or `recomputed_priority` to `new_pri`? We also have `num_con` and `neigh_pri` kicking around.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    return next(hub_itr, jump_to_dist);
+  } else {
+    //forwards label
+    auto first_fwd_bound_itr = next(start_itr, 1); 
+    auto last_fwd_bound_itr = next(start_itr, 1+node_count);
+    auto jump_to_dist = (*last_fwd_bound_itr) - *first_fwd_bound_itr;
+    return next(hub_itr, jump_to_dist); 
+  }
+}
+*/
```
adamnovak:
```suggestion
```

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+void down_dijk(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_back) {
+  auto in_node = node;
+
+  // TODO: We used to add -ov[node].seqlen to labels_back[node] for the hub
+  // ov[node].new_id. But this involved doing unsigned overflow shenanigans,
+  // and gave us values in the labels that are maximally wide and can't later
+  // be packed into the reduced bit width in a SnarlDistanceIndex.
+  //
+  // The tests didn't seem to cover a case where these entries were needed, so
+  // we just don't do that anymore.
```
adamnovak: We can probably cut this note now.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+  std::priority_queue<tuple<DIST_UINT, int>, vector<tuple<DIST_UINT, int>>, greater<tuple<DIST_UINT, int>>> q; 
+ 
+  auto [_, __] = out_edges(in_node, ov);
+  std::for_each(_, __, [&](auto edge) {
```
adamnovak: We should just replace this pattern everywhere with `std::ranges::for_each()`, which we should now have. I *think* Boost's idea of ranges can feed into that. I'm not sure if we need to plug in one of Boost's conversion functions.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+#ifdef debug_create
+    //if (j % 100 == 1) { 
+    cerr << j << "th node, " << v[j] << endl;
+#endif
+
+    //cerr << "starting dijkstra: " << endl;
+    down_dijk_back(v[j], ov, node_dists, labels, labels_back);
+  
+    //cerr << "starting rev dijkstra: " << endl;
+    down_dijk(v[j], ov, node_dists, labels, labels_back);
```
adamnovak: The commented-out code here should be removed.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    flat_label_vec[i+1+label_count+1] = current_offset; 
+    current_offset += label_size;
+
+    flat_vec_size += (label_size*2); 
+  }
+  flat_label_vec[(label_count+1)+label_count+1] = current_offset;      
+
+  flat_label_vec.resize(flat_vec_size);   
+
+  //copy info into vec (fwd)
```
adamnovak: This could be more specific

bdsg/src/landmark.cpp
```diff
@@ -0,0 +1,501 @@
...
+
+/*
+following functions assume input graph node ids go from 1...N 
+*/
+
+/*
+Converts a node_id to a nodeside_id
+*/
+//using bool for indicating side just like for direction in follow_edges
+NodesideId node_to_nodeside(NodeId node_id, bool left) {
```
adamnovak: I think the whole idea of nodesides isn't really used anymore.

bdsg/src/snarl_distance_index.cpp
```diff
@@ -480,25 +379,41 @@ net_handle_t SnarlDistanceIndex::get_parent(const net_handle_t& child) const {
...
+#ifdef debug_parent 
+    std::cerr << "Child " << net_handle_as_string(child) << " has simple snarl record type " << stringify(get_record_type(snarl_tree_records->at(get_record_offset(child)))) << " and current handle type " << stringify(get_handle_type(child)) << std::endl;
+#endif
+
         //If this is a simple snarl and a node or chain, then the parent offset doesn't change
         if (get_handle_type(child) == NODE_HANDLE) {
-            //If this is a node, then return it as a chain
+            // If this is a node, then return it as a chain
+            // TODO: Why can a simple snarl need to look like a node itself?
+            // TODO: Why can a simple snarl need to look like a chain? Because the node needs to look like a chain?
```
adamnovak: I added basically this same TODO several times, but we probably only need to ask the question once, where we would put the answer.

bdsg/src/snarl_distance_index.cpp
```diff
@@ -1120,11 +1054,20 @@ size_t SnarlDistanceIndex::distance_in_parent(const net_handle_t& parent,
...
+         << " and " << net_handle_as_string(child2)
+         << " facing back toward it in parent " << net_handle_as_string(canonical(parent)) << endl;
+    cerr << "\t\tChild parents are " << net_handle_as_string(canonical(child1_parent)) << " and " << net_handle_as_string(canonical(child2_parent)) << endl;
+
+    if (canonical(parent) != canonical(child1_parent) || canonical(parent) != canonical(child2_parent)) {
+        std::cerr << "Error: parent mismatch!" << std::endl;
+        std::cerr << as_integer(canonical(parent)) << " = " << net_handle_as_string(canonical(parent)) << std::endl;
+        std::cerr << as_integer(canonical(child1_parent)) << " = " << net_handle_as_string(canonical(child1_parent)) << std::endl;
+        std::cerr << as_integer(canonical(child2_parent)) << " = " << net_handle_as_string(canonical(child2_parent)) << std::endl;
+        assert(false);
```
adamnovak: This could be a `throw` and a `stringstream`.

bdsg/src/snarl_distance_index.cpp
```diff
@@ -1183,6 +1127,10 @@ size_t SnarlDistanceIndex::distance_in_parent(const net_handle_t& parent,
     } else if (is_chain(parent)) {
         if (get_record_handle_type(get_record_type(snarl_tree_records->at(get_record_offset(parent)))) == NODE_HANDLE ||
             get_record_handle_type(get_record_type(snarl_tree_records->at(get_record_offset(parent)))) == SNARL_HANDLE) {
+            // TODO: Why would this happen?
+#ifdef debug_distances
+        std::cerr << "=>They are not reachable because this chain is really a node or snarl(?!)" << std::endl;
+#endif
```
adamnovak: This section should be less shocked.

bdsg/src/snarl_distance_index.cpp
```diff
@@ -1317,93 +1269,145 @@ size_t SnarlDistanceIndex::distance_in_parent(const net_handle_t& parent,
...
+            //
+            // The HHL index thinks a child being "reverse" means that we're
+            // thinking of the child in the opposite orientation form how it
+            // appears in us. (So even not-reversed won't mean local forward
+            // orientation if is_reversed_in_parent() is true for that child).
+            //
+            // TODO: Probably need to also flip for is_reversed_in_parent() to
+            // account for this.
+            //
+            // TODO: dir1 and dir2 aren't just normal is_reverse flags.
```
adamnovak:
```suggestion
            // Note that dir1 and dir2 aren't just normal is_reverse flags. 
```

bdsg/src/snarl_distance_index.cpp
```diff
@@ -4186,44 +4135,60 @@ SnarlDistanceIndex::SnarlRecord::SnarlRecord (size_t pointer, const bdsg::yomo::
...
-        return vector_size;
-    } else if (type ==  OVERSIZED_SNARL){
-        //For a large min_distance snarl, all distances get stored in the children
-        return 0;
+    if (!is_nonsimple_snarl(type)) {
+        throw runtime_error("error: this is not a snarl");
+    }
+    if (has_distances(type)) {
+        if (is_oversized_snarl(type)) {
+            // For oversized snarls, vec_size for the hub labeling data is accounted separately.
```
adamnovak: See this is why `vec_size` is a weird name, because here we have to explain how it's not to be confused with or counted towards `distance_vector_size`.

bdsg/src/snarl_distance_index.cpp
```diff
@@ -4186,44 +4135,60 @@ SnarlDistanceIndex::SnarlRecord::SnarlRecord (size_t pointer, const bdsg::yomo::
...
-    } else if (type == DISTANCED_SNARL || type == DISTANCED_ROOT_SNARL) {
-        //For a normal min distance snarl just the distances between internal node sides
-        size_t node_side_count = node_count * 2;
-        size_t vector_size =  (((node_side_count+1)*node_side_count) / 2);
-        return vector_size;
-    } else if (type ==  OVERSIZED_SNARL){
-        //For a large min_distance snarl, all distances get stored in the children
-        return 0;
+    if (!is_nonsimple_snarl(type)) {
+        throw runtime_error("error: this is not a snarl");
```
adamnovak:
```suggestion
        throw runtime_error("error: trying to get size of distance matrix for something other than a snarl that would have one");
```

bdsg/src/snarl_distance_index.cpp
```diff
@@ -4186,44 +4135,60 @@ SnarlDistanceIndex::SnarlRecord::SnarlRecord (size_t pointer, const bdsg::yomo::
...
-size_t SnarlDistanceIndex::SnarlRecord::record_size (record_t type, size_t node_count) {
-    return SNARL_RECORD_SIZE + distance_vector_size(type, node_count);
+size_t SnarlDistanceIndex::SnarlRecord::record_size (record_t type, size_t node_count, size_t vec_size) { 
+    if (is_oversized_snarl(type)) { 
+      // Oversized snarls need the fixed-size header, the slot for the length
+      // of the packed hub label vector, and the packed hub label vector
+      // itself.
+      
+      // TODO: Can we stop storing the packed hub label vector length? Do we
+      // ever use it???
```
adamnovak: We should probably answer this question before we ship this; we can probably just stop storing it and see if tests break.

bdsg/src/snarl_distance_index.cpp
```diff
@@ -5165,37 +5184,60 @@ size_t SnarlDistanceIndex::ChainRecord::get_distance(size_t rank1, bool left_sid
...
+#ifdef debug_distances
+            std::cerr << "Distance from right to other right is " << distance << std::endl;
+#endif
         }
     } else if (left_side1 && left_side2) {
         //Left side of 1 and left side of 2
         if (rank1 == rank2) {
             distance = reverse_loop1;
+#ifdef debug_distances
+            std::cerr << "Distance on left slef loop is " << distance << std::endl;
```
adamnovak:
```suggestion
            std::cerr << "Distance on left self loop is " << distance << std::endl;
```

bdsg/src/snarl_distance_index.cpp
```diff
@@ -6383,20 +6406,19 @@ void SnarlDistanceIndex::get_snarl_tree_records(const vector<const TemporaryDist
...
+                            const TemporaryDistanceIndex::TemporarySnarlRecord& temp_snarl_record = temp_index->get_snarl(child_record_index);
                             if (!temp_snarl_record.is_trivial && !temp_snarl_record.is_simple) {
                                 //If this is an actual snarl that we need to make
 
 
                                 //Add the snarl to the chain, and get back the record to fill it in
 
                                 bool ignore_distances = (snarl_size_limit == 0) || only_top_level_chain_distances;
+                                
+                                record_t record_type = encode_nonroot_nonsimple_snarl(!ignore_distances, temp_snarl_record.is_regular, !ignore_distances && temp_snarl_record.node_count > snarl_size_limit);
```
adamnovak, line 6414: We might consider relaxing the very tight exception rules on the encoding function and making the last argument a don't-care when not relevant so we can avoid the `&&` here.

bdsg/src/test_libbdsg.cpp
```diff
@@ -4989,6 +4989,310 @@ void test_hash_graph() {
...
+      cerr << "(" << sz.hub << "," << sz.dist << ") "; 
+      } 
+      cerr << " | ";
+    }
+    cerr << endl;
+    cerr << "pack:" << endl; 
+    for (auto sz: packed_labels) {
+      cerr << sz << " ";
+    }
+    cerr << endl;  */
```
adamnovak: We should `#define` out or remove the commented-out code here.

bdsg/src/test_libbdsg.cpp
```diff
@@ -4989,6 +4989,310 @@ void test_hash_graph() {
...
+    // check u -> v and v -> u are different
+    // 3rd forward to 1st forward: shouldn't connect because nothing is downstream of 3rd
+    assert(hhl_query(packed_labels.begin(), rank(3, false), rank(1, false)) == INF_INT);
+    // 1st forward to 3rd forward: direct connection
+    assert(hhl_query(packed_labels.begin(), rank(1, false), rank(3, false)) == 0);
+
+    // need to debug
+    for (size_t a = 0; a < handles.size() * 2; a++) {
+      cerr << hhl_query(packed_labels.begin(), rank(1, false), a) << endl;
+    }
```
adamnovak: This has been debugged now, right? We should probably not print stuff from the tests unless they're built with a define.

bdsg/src/test_libbdsg.cpp
```diff
@@ -5089,6 +5393,7 @@ int main(void) {
     test_packed_subgraph_overlay();
     test_multithreaded_overlay_construction();
     test_mapped_packed_graph();
-    test_hash_graph();
-    test_snarl_distance_index();
+    test_hash_graph(); */
+    test_hub_labeling();
+    //test_snarl_distance_index();
```
adamnovak: We have to turn the other tests back on.

bdsg/src/vectorizable_overlays.cpp
```diff
@@ -173,16 +173,21 @@ void VectorizableOverlay::index_nodes_and_edges() {
         }
     }
     
+    // We limit threading on small inputs.
+    auto limited_threads = [&](size_t batch) {
+        return std::max<size_t>(1, std::min<size_t>(batch / 1024, get_thread_count()));
```
adamnovak: I think I added this, but this could use a constant instead of a magic number.

CMakeLists.txt
```diff
@@ -143,6 +143,8 @@ endif()
 # Find other system dependencies
 pkg_check_modules(Jansson REQUIRED IMPORTED_TARGET jansson)
 
+find_package(Boost REQUIRED)
```
adamnovak, line 146: If we're adding a dependency on Boost, we need to explain that in the README.

Makefile
```diff
@@ -24,6 +24,7 @@ OBJS += $(OBJ_DIR)/path_subgraph_overlay.o
 OBJS += $(OBJ_DIR)/subgraph_overlay.o
 OBJS += $(OBJ_DIR)/vectorizable_overlays.o 
 OBJS += $(OBJ_DIR)/packed_subgraph_overlay.o 
+OBJS += $(OBJ_DIR)/ch.o
```
adamnovak, line 27: So we don't actually need `hublabel.o` and `landmark.o` and we can drop their CPP files?

adamnovak (commented):
It looks like this brings along more of the other oracle approaches than we need to to implement just the contraction hierarchies and hierarchical hub labeling that actually gets used.

There's also some commented-out code that should be removed, and some documentation comments that ought to be written, some of which are for things I added.

Also, I didn't add a file comment for it, but if it's easy I might port away from Boost to our own data structure for what's now called the "Boost graph", to avoid the dependency. Right now we have Boost in vg only because of the old `Vowpal-Wabbit` MAPQ recalibration stuff in https://github.com/vgteam/vg/pull/1621 which I don't know actually has any users. If it's *not* easy it probably wouldn't be worth doing, or at least might not be worth doing now. (Since it's only needed for the construction, we can change later if we want to without invalidating indexes.)

bdsg/src/snarl_distance_index.cpp
adamnovak: OK, I considered it and I don't like it because we encode oversized-ness first.

bdsg/src/test_libbdsg.cpp
electricEpilith: pretty sure

Makefile
electricEpilith: yup, done

bdsg/src/test_libbdsg.cpp
electricEpilith: done

CMakeLists.txt
electricEpilith: done

bdsg/cmake_bindings/bdsg/snarl_distance_index.cpp
adamnovak, line 1: These files are auto-generated, so you *can* run the formatter on them, but the changed will be clobbered when we next regenerate them.

It might make sense to include the formatter script in the repo somewhere at some point, so we can keep around a command line that formats the real code and ignores the synthetic code.

bdsg/include/bdsg/ch.hpp
```diff
@@ -1,27 +1,72 @@
...
+  }
+  return (DIST_UINT)val;
+}
 
 /**
  * For a handle graph indexed with HHL, get the HHL rank ("Boost graph ID") for
  * an orientation of a node, as a source or destination.
  */
-NODE_UINT bgid(const handle_t& h, const bdsg::HashGraph& hg);
+NODE_UINT bgid(const handle_t &h, const bdsg::HashGraph &hg);
```
adamnovak, line 71: Is this a configurable switch of the formatter? We almost always use the ampersand on the type, and I don't think we want to go through and touch *all* the code to put it on the variable instead, so having the formatter come through and change it in *some* files is probably making things worse instead of better.

README.md
```diff
@@ -94,6 +94,8 @@ The documentation can then be found at `docs/_build/html/index.html`.
 - [`BBHash/alltypes`](https://github.com/rizkg/BBHash/tree/alltypes) 
 - [`jansson`](https://github.com/akheron/jansson)
 
+There is also a dependency on [`Boost`](https://www.boost.org/).
```
adamnovak: Well, yes, but we should mention that in the list like for the *other* dependencies.

adamnovak (commented):
The changes since my last review look pretty good, but there still seem to be a bunch of unaddressed comments from my last review, like missing include guards and commented-out code.

bdsg/cmake_bindings/bdsg/snarl_distance_index.cpp
electricEpilith: The formatter is `clang-format`, you'd have to install it on your own machine

bdsg/include/bdsg/ch.hpp
electricEpilith: added

bdsg/include/bdsg/ch.hpp
electricEpilith: Will need to look at `clang-format` docs

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: done

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: rejecting v4 indexes

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: done

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: yup

bdsg/src/ch.cpp
electricEpilith: ok

bdsg/src/ch.cpp
electricEpilith: removed

README.md
electricEpilith: ok

bdsg/src/ch.cpp
electricEpilith: done

bdsg/src/ch.cpp
electricEpilith: all occurrences resolved

bdsg/src/ch.cpp
electricEpilith: changed to `should_not_contract`

bdsg/src/ch.cpp
electricEpilith: done

bdsg/src/ch.cpp
electricEpilith: using `boost::make_iterator_range` now

bdsg/src/ch.cpp
electricEpilith: removed

bdsg/src/ch.cpp
electricEpilith: done

bdsg/src/ch.cpp
electricEpilith: ok

bdsg/src/ch.cpp
electricEpilith: removed vector

bdsg/src/ch.cpp
electricEpilith: fixed

bdsg/src/ch.cpp
electricEpilith: addressed

bdsg/src/ch.cpp
electricEpilith: renamed

bdsg/src/ch.cpp
electricEpilith: ok

bdsg/src/ch.cpp
electricEpilith: rewritten

bdsg/include/bdsg/internal/indexing_iterator.hpp
electricEpilith: added `operator[]`

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: renamed

electricEpilith: ok

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: done

bdsg/src/snarl_distance_index.cpp
electricEpilith: ok

bdsg/src/snarl_distance_index.cpp
electricEpilith: addressed

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,375 @@
...
+                   vector<vector<HubRecord>> &labels,
+                   vector<vector<HubRecord>> &labels_rev);
+
+void test_dijk(int node, CHOverlay &ov, vector<DIST_UINT> &node_dists,
+               vector<vector<HubRecord>> &labels,
+               vector<vector<HubRecord>> &labels_rev);
+
+void test_dijk_rev(int node, CHOverlay &ov, vector<DIST_UINT> &node_dists,
+                   vector<vector<HubRecord>> &labels,
+                   vector<vector<HubRecord>> &labels_rev);
```
adamnovak: These are testing functions that are also never called. They really ought to be in the testing code, or maybe they could be cut altogether.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1038 @@
...
+
+  while (!q.empty()) { node_dists[get<1>(q.top())] = INF_INT; q.pop(); } 
+}
+
+void down_dijk_back(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_back) {
+  auto in_node = node;
+  labels[node].emplace_back(ov[node].new_id, 0); 
+  node_dists[node] = 0;
+
+  //std::priority_queue<tuple<DIST_UINT, int>, vector<tuple<DIST_UINT, int>>, greater<tuple<DIST_UINT, int>>> q;
```
adamnovak:
```suggestion
```

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1038 @@
...
+  
+  node_dists[node] = INF_INT;
+  for (auto n: to_reset) { node_dists[n] = INF_INT; }
+
+  for (auto& t: q) { node_dists[get<1>(t)] = INF_INT; } 
+  q.clear(); 
+} 
+
+void test_dijk(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_back) {
+  auto in_node = node; //node_dists[node] = 0;
```
adamnovak:
```suggestion
  auto in_node = node;
```

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,375 @@
...
+  return dist;
+}
+
+void down_dijk(int node, CHOverlay &ov, vector<DIST_UINT> &node_dists,
+               vector<vector<HubRecord>> &labels,
+               vector<vector<HubRecord>> &labels_rev);
+
+void down_dijk_rev(int node, CHOverlay &ov, vector<DIST_UINT> &node_dists,
+                   vector<vector<HubRecord>> &labels,
+                   vector<vector<HubRecord>> &labels_rev);
```
adamnovak: These seem to be the core of label creation, but they don't have any comments explaining what they do, either here or in their implementations.

There should be doc comments here.

bdsg/include/bdsg/ch.hpp
```diff
@@ -0,0 +1,375 @@
...
+ */
+vector<size_t> pack_labels(const vector<vector<HubRecord>> &labels,
+                           const vector<vector<HubRecord>> &labels_back);
+
+// not necessary stuff
+void write_to_csv(CHOverlay &ov, string out_path);
+
+void write_to_gr(CHOverlay &ov, string out_path);
+
+vector<CHOverlay::vertex_descriptor> read_node_order(string in_path);
```
adamnovak: If these aren't necessary, they should be removed. If there's a good reason to keep them, they shouldn't be described as unnecessary.

bdsg/include/bdsg/ch.hpp
adamnovak, line 1: I had some review comments here about renaming stuff and trying to have a better notion of what is part of the interface of this component versus what is internal. All my comments became "outdated" when the file was reformatted.

We can drop those, but I still see some important functions in here that aren't documented at all, some test functions in here that aren't called and probably don't belong in here, and some functions that are supposedly not necessary but are in here anyway.

bdsg/src/ch.cpp
```diff
@@ -0,0 +1,1038 @@
...
+  // edge-difference calculations.
+  //
+  // contracted_neighbors and level are part of the priority formula used when
+  // choosing which node to contract next: nodes adjacent to many already-
+  // contracted nodes (or near high-level nodes) are deprioritised so we build
+  // a good hierarchy.
+  //
+  // should_not_contract[neighbour] = true prevents a neighbour from being
+  // contracted in the same "round", ensuring the independence heuristic that
+  // keeps the contraction order sensible.
```
adamnovak: This doesn't make a lot of sense to me as a phrase. I guess the idea is that if we operate on neighboring nodes it violates some assumption, right?

But we haven't introduced an "independence heuristic" in the explanation, and I don't know what it means to "ensure" a heuristic. This sounds like it's written up in the form of a concluding sentence when the content is something else, and I find reading it baffling.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -199,20 +199,24 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
     void serialize_members(std::ostream& out) const;
     void deserialize_members(std::istream& in);
 
-    /// Call when loading a distance index; will error if wrong version
+    /* Call when loading a distance index; will error if wrong version
```
adamnovak: These comments being touched here all need 2 leading stars (`/** ... *./`) if we still want them to be picked up as doc comments by the documentation generators.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -610,51 +627,255 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
-            || type == DISTANCED_SNARL || type == OVERSIZED_SNARL || type == DISTANCED_ROOT_SNARL 
+                   CHILDREN,
+                   OVERSIZED_ROOT_SNARL};
+
+    // Because the record_t encodes a complex taxonomy of snarls not *quite*
+    // decomposable to flags, we use these accessors to look at facets of it.
+
+    /*Return true if records of the given type have stored distances.
+     *
+     */
```
adamnovak: These new doc comments also need 2 leading stars. And I would set them out like this:


```suggestion
    /**
     * Return true if records of the given type have stored distances.
     */
```

See https://www.doxygen.nl/manual/docblocks.html (which says the old `///` way I had them isn't actually right either).

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1605,32 +1887,74 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+            /* will be std::numeric_limits<size_t>::max(). If any lopp is
+     *
+     */
+            /* possible anywhere along the chain, the last entry will contain
+     *
+     */
+            /* a possible loop distance.
+     *
+     */
```
adamnovak: There's something wrong with the comment formatting here.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1605,32 +1887,74 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+            /* snarl, not including boundary nodes.
+     *
+     */
             size_t distance_start_start = std::numeric_limits<size_t>::max();
+            /* Minimum distance from the end back to itself within the snarl,
+     *
+     */
+            /* not including boundary nodes.
+     *
+     */
```
adamnovak: These comments aren't laid out right.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1642,19 +1966,28 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+            // The end node is reverse if it points back into the snarl, not along it.
             bool end_node_rev;
             bool is_trivial;
             bool is_simple;
+            /* Set to true if the snarl is regular (see SnarlDistanceIndex::is_regular_snarl()).
+     *
+     */
+            /* If is_simple is true, this must also be set to true when filling in the TemporarySnarlRecord.
+     *
+     */
```
adamnovak: This comment also isn't laid out right.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1683,12 +2016,122 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+        
+        /* Look up a snarl from a temporary record reference.
+     *
+     */
+        /* Throws an error if the reference is not to a snarl or is out of bounds.
+     *
+     */
+        /* This version can be used when the object is const.
+     *
+     */
```
adamnovak: These comments also aren't laid out right.

adamnovak (changes requested):
This looks pretty good, but there are still some fairly big functions that are neither called nor documented (and appear to be removable), and it looks like something went wrong with an attempt to add doc comments by find-and-replace which probably needs to be cleaned up.

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: fixed

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: fixed

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: ok

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: ok

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: ok

bdsg/include/bdsg/ch.hpp
electricEpilith: added doc comments

bdsg/include/bdsg/ch.hpp
electricEpilith: removed

bdsg/include/bdsg/ch.hpp
electricEpilith: removed

bdsg/include/bdsg/snarl_distance_index.hpp
electricEpilith: done

bdsg/src/ch.cpp
electricEpilith: edited comment to be more clear

bdsg/include/bdsg/ch.hpp
electricEpilith: think I've addressed these now?

bdsg/include/bdsg/ch.hpp
```diff
@@ -328,22 +328,50 @@ DIST_UINT hhl_query(ItrType start_itr, size_t rank1, size_t rank2) {
...
+ * Run a pruned forward Dijkstra from node over the contraction hierarchy ov,
+ * contributing node (by its CH rank, ov[node].new_id) as a hub to the backward
+ * (in-) labels of the nodes it can reach.
+ *
+ * For each node reachable from node along forward (out-) edges, the candidate
+ * distance is compared against what the already-built labels can already
+ * express (binary_intersection_ch of labels_back[cur] against labels[node]).
+ * If the existing labels do not already cover the distance, node's rank is
+ * appended as a hub to labels_back[cur] and the search continues from cur;
+ * otherwise that branch is pruned.
```
adamnovak, line 341: This is a summary of how the implementation currently works (which would be better in the function body), not documentation for what the function promises to accomplish or requires in order to do it.

bdsg/include/bdsg/ch.hpp
```diff
@@ -328,22 +328,50 @@ DIST_UINT hhl_query(ItrType start_itr, size_t rank1, size_t rank2) {
...
+ * Run a pruned backward Dijkstra from node over the contraction hierarchy graph ov,
+ * contributing node (by its CH rank, ov[node].new_id) as a hub to the forward
+ * (out-) labels of the nodes that can reach it.
+ *
+ * Seeds labels[node] with node itself at distance 0, then explores backward
+ * (in-) edges. For each node from which node is reachable, the candidate
+ * distance is compared against what the already-built labels can already
+ * express (binary_intersection_ch of labels[cur] against labels_back[node]).
+ * If not already covered, node's rank is appended as a hub to labels[cur] and
+ * the search continues; otherwise that branch is pruned.
```
adamnovak, line 362: This also seems more like a summary.

bdsg/include/bdsg/snarl_distance_index.hpp
```diff
@@ -1021,16 +997,15 @@ class SnarlDistanceIndex : public SnarlDecomposition, public TriviallySerializab
...
+/** 
  * These are used to interpret snarl_tree_records, which is just a vector of ints.
  * Uses a SnarlTreeRecord as the main class for defining and interpreting the records.
  * SnarlTreeRecords are given a base pointer, which points to the start of the record
  * Each record starts with a "tag", which defines which type of record it is. The 
  * contents of the record are defined below: 
-*/
-    
-    /*Root record
+ */
```
adamnovak, line 1006: This doesn't apply to a particular member after the comment, and so can't really be a doc comment.

bdsg/include/bdsg/snarl_distance_index.hpp
adamnovak, line 1014: This is going to get used as the doc comment for just the first variable. It's really about the whole group of them, but these pseudo-structs are not a thing Doxygen lets you document.

I guess that makes as much sense as anywhere else to put the text in the docs?

adamnovak, line 1215: I think these both are going to end up as the documentation for `get_record_type()`.

adamnovak, line 1716: This block about "Functions to add children to a chain." is the doc comment for `add_snarl`, and the line about what `add_snarl` in particular does gets thrown away.

bdsg/src/ch.cpp
```diff
@@ -727,10 +741,15 @@ DIST_UINT binary_intersection_ch(vector<HubRecord>& v1, vector<HubRecord>& v2) {
...
 
 
-void down_dijk(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_back) {
+/** Pruned forward Dijkstra from node, appending node as a hub to the backward
+ * labels (labels_back) of the nodes it reaches. See build_backward_labels() in
+ * ch.hpp. The check_dist comparison prunes branches whose distance the
+ * existing labels already cover, and node_dists is left reset to INF_INT.
+ */
+void build_backward_labels(int node, CHOverlay& ov, vector<DIST_UINT>& node_dists, vector<vector<HubRecord>>& labels, vector<vector<HubRecord>>& labels_back) {
   auto in_node = node;
```
adamnovak, line 750: I don't know what Doxygen will do if we have two doc comments for the same function in different places.

But since we're not binding ch.hpp they won't end up on the web page Python docs, so maybe we don't care as much?

adamnovak (approved):
I still think the doc comment syntax isn't all quite right, but this is probably good enough to ship!

PR #241, opened 2026-05-26
Make position indexing work more similarly across implementations

adamnovak:
This makes `PositionOverlay`, `MutablePositionOverlay`, `ReferencePathOverlay`, `PackedPositionOverlay`, and `PackedReferencePathOverlay` work more similarly.

Now they all look at REFERENCE and GENERIC-sense paths by default, and can be told to look at other paths in a similar way.

Some of them completely hide the un-indexed paths, and others raise exceptions when trying to do operations on paths that would need the index when they aren't indexed, but both of those behaviors are now under test.

This is related to #236 and #237.


adamnovak:
I need to regenerate the bindings in order to merge this.

